PR #85, opened 2018-05-22
defer multiplying out sums until they are summed

avi-stripe:
We've previously had an optimization where, given a multiplication of the form `(ax + b)(cy + d)` (where everything but `x` and `y` are constants) we would expand it out to `(acxy + bcy + adx + bd)`, in the hopes that we might later on sum this result with something else that had `x` and `xy` terms and the constants would combine. Very recently, we generalized this to multiplications of sums with more than two terms (within limits), which helped specifically with the "full normal" benchmark (fitting a normal that has uniform priors for mean and sd).

This PR stems from the observation that the only real reason to expand out these multiplications is when we are immediately going to sum the results with something else; otherwise we should keep it in product form; either we'll do something non-linear with the product and foreclose usefully multiplying it out (in which case we're glad we didn't do that for nothing), or we'll keep accumulating more terms into the product, which doesn't stop us from multiplying the whole thing out later; or we will, indeed, try to sum something with it, in which case we can take advantage of the distribution rule with more hope that it will be useful, and with an entire multi-term product at once.

That is, what we are doing here is:
* assume we have some expression `s = y_0^k_0 * y_1^k_1 * ...`, where `k_n` is constant
* assume that the `y_n` terms in `s` are of themselves of the form `a_n_0 * x_n_0 + a_n_1 * x_n_1 + ...` where `a_n_m` is constant
* assume we are now trying to construct `s + t` for some unknown `t`
* we want to restructure `s` to end up as something like `((a_0_0 * a_1_0 * ...) * (x_0_0 * x_1_0 * ...)) + ...)`, up to some maximum number of additive terms
* we want to do this because we hope that the non-constant `(x_0_0...)` parts of these terms may be in common with something in `t` (or some later thing we will add `s+t` to), and we can combine the constants

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/LogLineOps.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/LogLineOps.scala
@@ -9,7 +9,7 @@ private[compute] object LogLineOps {
     if (merged.isEmpty)
       Real.one
     else
-      LogLine(LineOps.merge(left.ax, right.ax))
+      LogLine(merged)
   }
 
   def pow(line: LogLine, v: Double): LogLine =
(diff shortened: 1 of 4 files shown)


gkk-stripe:
the idea looks solid to me, posted some questions

rainier-core/src/main/scala/com/stripe/rainier/compute/LogLineOps.scala
```diff
@@ -24,6 +24,58 @@ private[compute] object LogLineOps {
    */
   def log(@unused line: LogLine): Option[Real] = None
 
+  /*
+  If possible, return a representation of ax as a list of terms to be summed. Within limits, it's helpful here
+  to distribute the multiplications across any internal sums, because that may well surface some terms
+  that are in common with whatever we're about to sum this with.
+   */
+  val DistributeToMaxTerms = 20
+  def distribute(line: LogLine): Option[Line] = {
```
gkk-stripe: what's the O complexity of `distribute`? is it still true that `LogLine` is small?

rainier-core/src/main/scala/com/stripe/rainier/compute/LogLineOps.scala
```diff
@@ -24,6 +24,58 @@ private[compute] object LogLineOps {
...
+        l.ax.size
+      else
+        l.ax.size + 1
+    def nTerms2(l: Line): Int = {
+      val n = nTerms(l)
+      (n * (n + 1)) / 2
+    }
+
+    val initial = (List.empty[(NonConstant, Double)], Option.empty[Line])
+    val (factors, terms) = line.ax.foldLeft(initial) {
```
gkk-stripe: have you considered using recursion instead of foldLeft? i think it would be easier to read; especially the condition when you bail out for optimizations would be more direct

avi-stripe: I agree that this is kinda hard to read but I'm not sure how recursion would help, it's structurally a pretty straightforward foldLeft. (Can you explain what you mean by "bail out"?Which line?)

rainier-core/src/main/scala/com/stripe/rainier/compute/LogLineOps.scala
avi-stripe: It's O(1), and yes, LogLine should still be small.

avi-stripe: Gah. I meant to write O(n).

rainier-core/src/main/scala/com/stripe/rainier/compute/LogLineOps.scala
gkk-stripe: line 57, i thought this was a fall-through condition that, once reached, will be exercised for the rest of foldLeft but i think i read this wrong
you could hit that pattern and then still hit other cases for subsequent values of ax, right?

avi-stripe: that's right.

gkk-stripe:
approved but just came up as thought: move "That is, what we are doing here is:" comment into `distribute`'s comments?

avi-stripe:
@gkk-stripe good idea, will do.

roban-stripe:
Post-merge review: LGTM :)

PR #89, opened 2018-05-23
stable RNG seed for tests

avi-stripe:
Report on the chosen seed whenever we create a new RNG, and use a chosen seed that makes the test pass so they're not flaky.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/sampler/RNG.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/sampler/RNG.scala
@@ -10,7 +10,11 @@ trait RNG {
 }
 
 object RNG {
-  val default: RNG = new ScalaRNG(System.currentTimeMillis)
+  lazy val default: RNG = {
+    val seed = System.currentTimeMillis
+    println("Initializing RNG with seed " + seed)
+    ScalaRNG(seed)
+  }
 }
 
 final case class ScalaRNG(seed: Long) extends RNG {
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/sampler/RNG.scala
```diff
@@ -10,7 +10,11 @@ trait RNG {
 }
 
 object RNG {
-  val default: RNG = new ScalaRNG(System.currentTimeMillis)
+  lazy val default: RNG = {
+    val seed = System.currentTimeMillis
+    println("Initializing RNG with seed " + seed)
```
andyscott-stripe: How about we remove the `println` from the main lib code (to avoid noise for folks using the lib) and:
- assign the seed to a publicly accessible `lazy val`
- update the tests to print the seed before running

andyscott-stripe: Also: Do we need the ability to set the seed before running? It might be worth reading an env var like `STRIPE_RAINIER_RNG_SEED` and using the `currentTimeMillis` as a fallback.

avi-stripe: The use case I'm thinking about is the one where someone runs their code and sees some funky results that they want to reproduce. They won't know in advance that they need to capture the seed, and by the time they see the results it'll be too late to. On the other hand I don't mind them having to change their code (vs ENV) in order to select a static seed.

avi-stripe: @andyscott-stripe if you have other thoughts on how to recover a seed from a failing output without causing noise I'm all ears (eg it could go in a tmp file but that feels awkward...)

andyscott-stripe: tmp file definitely seems awkward!

PR #91, opened 2018-05-23
pointer to maven artifact in README

avi-stripe:
what it says on the tin


avi-stripe:
going to self-approve since this is trivial and the test failure will go away once it's on master

travisbrown-stripe:
👍

PR #92, opened 2018-05-23
docs on real

avi-stripe:
This adds a tour-style doc of `Real`. It also makes some minor tweaks to `LineOps` to make the demo a little nicer (in particular, backing out to the original object when we just have `1x + 0`).


gkk-stripe:
lgtm

awesome docs!

andyscott:
Nice!

roban-stripe:
This is so, so useful (like the rest of the docs)! Thanks for adding this!

Issue #93, opened 2018-05-24
Bug in Binomial likelihood?


darrenjw:
Obviously I could be misunderstanding something, but I wonder if there might be a bug in the Binomial logDensity. It isn't behaving as I would expect. A script to illustrate the problem is appended below. I would expect the posterior to be centred on the true value of 0.2.

```scala
import com.stripe.rainier.compute._
import com.stripe.rainier.core._
import com.stripe.rainier.sampler._
import com.stripe.rainier.repl._
import scala.util.Random

object BinomialTest {

  def genBernoulli(p: Double, r: Random): Int =
    if (r.nextDouble < p) 1 else 0

  def genBinomial(n: Int, p: Double, r: Random): Int =
    (1 to n).map(i => genBernoulli(p,r)).sum

  def main(args: Array[String]): Unit = {

    // first simulate some data
    val r = new Random
    val N = 10000
    val n = 10
    val p = 0.2
    val x = (1 to N) map { i => genBinomial(n,p,r) }
    println(x.take(20))

    // now build and fit model
    val model = for {
      p <- Uniform(0, 1).param
      _ <- Binomial(p,n).fit(x)
    } yield (p)

    implicit val rng = RNG.default

    val out = model.sample()
    //val out = model.sample(HMC(5), 1000, 10000)
    println(out.take(10))
    println(s"p (true value $p):")
    println(DensityPlot().plot1D(out).mkString("\n"))

  }

}
```

avi-stripe:
It's entirely likely that there's a bug since the discrete distributions have no test coverage. 😢 I'll start working on a test case similar to `ContinuousTest` for this. Thanks for surfacing.

PR #94, opened 2018-05-24
fix bug with Binomial

avi-stripe:
The `logDensity` of the convenience `Binomial` wrapper around `Multinomial` was not correctly filling in the `false` side of the (true,false) counts vector it was passing through. This PR should fix https://github.com/stripe/rainier/issues/93


darrenjw:
Great, thanks - that seems correct. Looks good to merge.

PR #95, opened 2018-05-25
Add effective sample size diagnostics

avi-stripe:
This adapts the previous `checkConvergence` into a `sampleWithDiagnostics` method which returns a `Diagnostics` object containing both `rHat` and `effectiveSampleSize` for each raw parameter. This is based on sections 30.3 and 30.4 in the Stan manual (including a correction to the final Neff equation in 30.4 that hasn't been published yet).

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
@@ -54,29 +54,29 @@ class RandomVariable[+T](private val value: T,
       .map(_.head.apply())
       .toList
 
-  def checkConvergence[V](sampler: Sampler,
-                          warmupIterations: Int,
-                          iterations: Int,
-                          chains: Int)(
+  def sampleWithDiagnostics[V](sampler: Sampler,
+                               warmupIterations: Int,
+                               iterations: Int,
+                               chains: Int)(
       implicit rng: RNG,
-      sampleable: Sampleable[T, V]): (List[V], List[Double]) = {
+      sampleable: Sampleable[T, V]): (List[V], List[Diagnostics]) = {
     val fn = sampleable.prepare(value, density.variables)
     val samples = 1.to(chains).par.map { _ =>
       sampler
         .sample(density, warmupIterations)
         .take(iterations)
         .map { s =>
-          fn(s.parameters)
+          (s.parameters, fn(s.parameters))
         }
         .toList
     }
     val allSamples = samples.toList.flatMap { chain =>
       chain.map(_._2)
     }
-    val rHats = Sampler.gelmanRubin(samples.toList.map { chain =>
+    val diagnostics = Sampler.diagnostics(samples.toList.map { chain =>
       chain.map(_._1)
     })
-    (allSamples, rHats)
+    (allSamples, diagnostics)
   }
 
   def toStream[V](sampler: Sampler, warmupIterations: Int)(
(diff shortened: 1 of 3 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -54,29 +54,29 @@ class RandomVariable[+T](private val value: T,
...
-                          warmupIterations: Int,
-                          iterations: Int,
-                          chains: Int)(
+  def sampleWithDiagnostics[V](sampler: Sampler,
+                               warmupIterations: Int,
+                               iterations: Int,
+                               chains: Int)(
       implicit rng: RNG,
-      sampleable: Sampleable[T, V]): (List[V], List[Double]) = {
+      sampleable: Sampleable[T, V]): (List[V], List[Diagnostics]) = {
```
andyscott: I'd encourage return types on all public methods. If you run the `scalafix` command it will attempt to add them for you and then you can clean them up by hand.

avi-stripe: agreed, but where was this comment meant for? this method has a return type

rainier-core/src/main/scala/com/stripe/rainier/sampler/Sampler.scala
```diff
@@ -1,13 +1,85 @@
...
+                      n: Double,
+                      m: Double,
+                      v: Double,
+                      lag: Int,
+                      acc: Double): Double = {
+    val vt = traces.map { trace =>
+      variogram(trace, lag)
+    }.sum / m
+    val pt = 1.0 - (vt / (2.0 * v))
+    if (pt > 0.0 && lag < 100)
```
avi-stripe: This probably isn't the right termination criteria. The Stan manual mentions that NUTS never produces negative autocorrelation and so they use it as an indication that noise > signal, but since we're using HMC not NUTS this likely doesn't apply. Not sure what I should use, though.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
andyscott-stripe: Apparently I was having a hard time reading when I wrote this original comment. (Maybe writing too much Java/Bash/Go/Python lately has ruined me for Scala.)

rainier-core/src/main/scala/com/stripe/rainier/sampler/Sampler.scala
```diff
@@ -1,13 +1,85 @@
...
+  def apply(traces: Seq[Array[Double]]): Diagnostics = {
+    val m = traces.size.toDouble
+    val n = traces.head.size.toDouble
+    val (rHat, v) = rHatAndV(traces, n, m)
+    val ac = autocorrelation(traces, n, m, v, 1, 0.0)
+    val ess = n * m / (1 + (2 * ac))
+    Diagnostics(rHat, ess)
+  }
+
+  def rHatAndV(traces: Seq[Array[Double]],
```
stephanie-stripe: might be helpful to reference the stan manual sections in comments here (or at the object level)

roban-stripe: ==

stephanie-stripe:
I don't think I can comment on the termination condition, but other than that (and the comment on documentation), this change looks good to me.

avi-stripe:
Thanks - I'll add the comments and merge, but keep looking around for what make a better termination condition for us.

PR #100, opened 2018-05-28
Sample EvilPlot integration

hibikir:
Ascii, CLI plots are nice and pretty, but some might prefer some pngs: EvilPlot can do that.

Evilplot's contours don't quite use the same algorithm as your density plots, so there are some tradeoffs: I'd argue that the large majority of plots are better, while the funnel plot doesn't show a funnel at all. Therefore, adding a new chart type that is closer to the ascii chart might be advisable.

As far as the code goes, there's very little to it: There's about 3 real statements, and the rest is just sbt changes and making the samples also output the pngs, just to show that we can. I have no special love for any of it, so feel free to ask for changes, or just make changes on top.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,6 +4,7 @@ crossScalaVersions in ThisBuild := List("2.11.12", "2.12.4")
 
 val unpublished = Seq(publish := {}, publishLocal := {}, publishArtifact := false)
 
+
 val publishSettings = Seq(
   releaseCrossBuild := true,
   releasePublishArtifactsAction := PgpKeys.publishSigned.value,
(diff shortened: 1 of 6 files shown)


hibikir:
travis seems unable/unwilling to find evilplot, but it finds it just fine in my `sbt test` run :?

avi-stripe:
That's weird. I don't have much travis experience. Maybe @andyscott-stripe has thoughts?

andyscott-stripe:
Seems reasonable to me 👍.

avi-stripe:
I'm going to close this in favor of https://github.com/stripe/rainier/pull/106 . However, I think a simpler `plotContour` API like this has (vs the full trace/acf/pairs stuff) would still be valuable, as would getting 2.11 artifacts for evilplot.

Issue #96, opened 2018-05-25
HMC sampler sometimes hangs on long runs


darrenjw:
I've been having a bit of trouble with the HMC sampler sometimes hanging on long runs. When monitoring CPU usage, a normal running chain usually hovers around a bit more than 100% (linux, java 8). But sometimes after a few minutes the usage will jump up to over 300%. Whenever this happens, the sampler never completes. I've managed to construct a reasonably "minimal" example which reliably fails for me. It should take less than 10 minutes to complete on a half-decent machine. If you drop the value of "thin" down from 1000 to 100 it completes fine in under a minute. But as-is, the CPU usage jumps after a few minutes and the job never completes.

```scala
import com.stripe.rainier.compute._
import com.stripe.rainier.core._
import com.stripe.rainier.sampler._
import com.stripe.rainier.repl._

object Failing {

  def main(args: Array[String]): Unit = {

    // first simulate some data from an ANOVA model
    val r = new scala.util.Random(0)
    val n = 50 // groups
    val N = 250 // obs per group
    val mu = 5.0 // overall mean
    val sigE = 2.0 // random effect SD
    val sigD = 3.0 // obs SD
    val effects = Vector.fill(n)(sigE * r.nextGaussian)
    val data = effects map (e => Vector.fill(N)(mu + e + sigD * r.nextGaussian))

    // build and fit model
    val prior = for {
      mu <- Normal(0, 100).param
      sigD <- LogNormal(0, 10).param
      sigE <- LogNormal(1, 5).param
    } yield Map("Mu" -> mu, "sigD" -> sigD, "sigE" -> sigE)

    def addGroup(current: Map[String, Real], i: Int) = for {
        gm <- Normal(current("Mu"), current("sigE")).param
        _ <- Normal(gm, current("sigD")).fit(data(i))
      } yield gm 

    val model = for {
      current <- prior
      _ <- RandomVariable.traverse((0 until n) map (addGroup(current, _)))
    } yield current

    implicit val rng = ScalaRNG(3)
    println("Model built. Sampling now...")
    val its = 10000
    val thin = 1000
    val out = model.sample(HMC(5), 1000000, its*thin, thin)
    println("Sampling finished.")

  }

}
```

avi-stripe:
The CPU jumping to 300% sounds like a GC problem. Thank you for the minimal repro, I'll take a look.

PR #101, opened 2018-05-29
No allocation in HMC

avi-stripe:
This is a major refactor of the `HMC` implementation to make it allocate much, much less. In particular, previously it would allocate on the order of `k*n` arrays, where `k` is the number of steps per HMC iteration and `n` is the number of iterations; now it allocates on the order of `n/t` arrays, where `t` is the thinning factor; in other words it only allocates arrays for the parameter vectors you actually want to keep. This lets you run models for large numbers of iterations without freaking out the garbage collector as in https://github.com/stripe/rainier/issues/96 .

Doing this required:
* changing the `ir` code generation to expect output and globals arrays to be passed in, rather than allocating them in the generated bytecode. Although the `Compiler` preserves the old, more convenient API, there's now also a `compileUnsafe` option which returns a raw `ir.CompiledFunction` object you can use if you want to pass in your own arrays, and which provides the necessary methods for querying the expected sizes of these arrays.
* rolling back the `toStream` sampling API, and instead pushing knowledge of number of iterations and thinning factor back into the `Sampler` interface; this allows individual samplers to more carefully control allocation.
* making relatively minor changes to `DualAvg` to shift it from an immutable to a mutable implementation
* making major changes to `HMC`: the immutable `HParams` and `HamiltonianChain` objects have been deleted, and their logic moved into `LeapFrog`; `LeapFrog` is still logically stateless, but now keeps a number of reusable buffers (and so is not threadsafe); the chain state is just kept by `HMC` and `DualAvg` in a single, opaque `Array[Double]`.


avi-stripe:
@darrenjw aha, thanks, I had only verified that it passed the existing tests but did not look at the output of your model at all. Any further details or examples would be helpful.

avi-stripe:
I've identified and fixed one bug which was causing the `NaN`s in that second example, but it's still doing wonky things on the original example. Flipping this back to `WIP` while I figure it out.

avi-stripe:
@darrenjw ok, this should be in good shape now. Please let me know what you think.

PR #104, opened 2018-05-30
Guard fatal warnings with a flag and only enable by default in CI

andyscott:
I think this should be a reasonable solution for the problem of local development workflow. Fatal warnings are now off by default for local development, but on by default for CI.

Either set `SCALAC_FATAL_WARNINGS=true` before running SBT or enter `set scalacFatalWarnings := true` from the SBT prompt to enable fatal warnings for local development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,31 +54,41 @@ If you're more familiar with deep learning systems like TensorFlow or PyTorch, y
 Rainier uses [SBT](https://www.scala-sbt.org/) to build. If you have SBT installed, you can build Rainier and test that it's working by executing `sbt "project rainierExample" run` and then selecting `rainier.example.FitNormal`. You should see output something like this:
 
 ```
-[info] Running rainier.example.FitNormal 
-    2.19 |                                                                                
-         |                                 ·           ··                                 
-         |               ·           ·  · ·· ·  ···  ·  ·· ·  ·    · · ·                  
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,32 +54,42 @@ If you're more familiar with deep learning systems like TensorFlow or PyTorch, y
...
+         |           ··················∘∘∘∘∘∘○○○○○○○○○○○○○∘∘∘∘∘∘···················
+         |         ·····················∘∘∘∘∘∘∘○○○○○∘∘○○∘∘∘∘∘∘∘∘················ ·····
+         |          ·······················∘∘∘∘∘∘∘∘∘∘∘∘∘∘∘∘······················· ··  ·
          |       ·· ··························································   ·  ·    ·
-    1.90 |          ·  ························································     ·     
-         |                ············································· ···· ·    · ·     
-         |                      ·   · ··· ··················· ···· ·  ·    ·              
-         |                           ··· ·  · ·· ·· ·  ··  · ·· ··                        
-         |                              ·        ··                                       
-    1.81 |                                       ·
```
andyscott: Emacs had some fun deleting trailing whitespace. I can update the PR for a minimal diff, if desired.

avi-stripe:
Nice. LGTM.

Issue #102, opened 2018-05-29
Richer set of examples


avi-stripe:
It would be wonderful to have some more sophisticated models as examples than the toy ones that are currently there. @darrenjw, I believe you have some - would it make sense to contribute them to the repo?

darrenjw:
Yes - I was thinking the same. I'll add a few ASAP.

darrenjw:
I'm tidying up a few examples to contribute, but I've been using EvilPlot to visualise the output (high-res trace plots are still the best MCMC convergence diagnostic I know of...), which will presumably break travis. To do this I have developed some code for EvilPlot integration (addressing #50), which is a bit more sophisticated than the simple contour plot in the outstanding pull request.

Would you like me to contribute the examples with the EvilPlot code commented out, with a view to commenting back in when travis is happy with the added dependency, or would you prefer me to remove all traces of the EvilPlot code for now? Or just wait until the EvilPlot dependency is sorted before adding the examples?

avi-stripe:
I think the best approach would be:

* put your `EvilPlot` code in a separate `rainier-plot` module which only builds under scala 2.12
* have `rainier-example` depend on `rainier-plot` so that you can use EvilPlot from there without commenting anything out

I don't think it's a problem for the examples to require 2.12 because nobody will need to include the example module as an artifact in their own projects.

avi-stripe:
(If you need help with the separate modules and restricting the build to 2.12 I'm sure we can do that part)

darrenjw:
We should perhaps leave this open for now. It would be good to have a bunch more examples. I've got a couple in the works, but still having some issues with sampling.

Issue #50, opened 2018-05-10
Add integration with EvilPlot


avi-stripe:
https://github.com/cibotech/evilplot

Just in case someone isn't hipster enough to appreciate my ASCII plots.

hibikir:
How good do you want the integration to be? If all you want is to be able to generate some contours, I have a 5 line implementation in my fork that I'll be happy to contribute. This is the bonus file output of one of the examples.
![plot](https://user-images.githubusercontent.com/7410363/40594114-ae233024-61f2-11e8-9990-62e41dd5bf76.png)

avi-stripe:
By all means, love to have the contribution.

darrenjw:
I think we can probably close this one.

PR #106, opened 2018-05-31
Integration with EvilPlot and two new examples using the plot package

darrenjw:
This PR addresses #102 and #50. As suggested in #102, I've created a new plot project containing the evilplot code and evilplot dependency. I've then made the examples project depend on the plot project so I can use the plot code in the examples. It all seems to work OK.

The bit I don't know how to do is stop the cross-building of some projects to 2.11, so I guess that travis will fail. However, I'm sure there's someone at Stripe who knows how to fix that...

### DIFF
--- a/rainier-example/src/main/scala/com/stripe/rainier/example/Anova.scala
+++ b/rainier-example/src/main/scala/com/stripe/rainier/example/Anova.scala
@@ -1,0 +1,92 @@
+/*
+Anova.scala
+
+Try doing a one-way ANOVA with random effects model using Rainier
+
+ */
+
+package com.stripe.rainier.example
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.core._
+import com.stripe.rainier.sampler._
+import com.stripe.rainier.repl._
+
+object Anova {
+
+  def main(args: Array[String]): Unit = {
+
+    // first simulate some data from an ANOVA model
+    val r = new scala.util.Random(0)
+    val n = 50 // groups
+    val N = 250 // obs per group
+    val mu = 5.0 // overall mean
+    val sigE = 2.0 // random effect SD
+    val sigD = 3.0 // obs SD
+    val effects = Vector.fill(n)(sigE * r.nextGaussian)
+    val data = effects map (e => Vector.fill(N)(mu + e + sigD * r.nextGaussian))
+
+    // build model
+    val prior = for {
+      mu <- Normal(0, 100).param
+      sigD <- LogNormal(0, 10).param
+      sigE <- LogNormal(1, 5).param
+    } yield Map("Mu" -> mu, "sigD" -> sigD, "sigE" -> sigE)
+
+    def addGroup(current: Map[String, Real], i: Int) =
+      for {
+        gm <- Normal(current("Mu"), current("sigE")).param
+        _ <- Normal(gm, current("sigD")).fit(data(i))
+      } yield gm
+
+    val model = for {
+      current <- prior
+      _ <- RandomVariable.traverse((0 until n) map (addGroup(current, _)))
+    } yield current
+
+    // fit model
+    implicit val rng = ScalaRNG(3)
+    println("Model built. Sampling now (will take a long time)...")
+    val its = 10000
+    val thin = 10000
+    val out = model.sample(HMC(5), 1500000, its * thin, thin)
+    println("Sampling finished.")
+
+    // process output
+    // first some ASCII plots
+    println(s"mu (true value $mu):")
+    println(DensityPlot().plot1D(out map (_("Mu"))).mkString("\n"))
+    println(s"sigE (true value $sigE):")
+    println(DensityPlot().plot1D(out map (_("sigE"))).mkString("\n"))
+    println(s"sigD (true value $sigD):")
+    println(DensityPlot().plot1D(out map (_("sigD"))).mkString("\n"))
+
+    // now some EvilPlots
+    import com.cibo.evilplot.plot._
+    import com.cibo.evilplot.geometry.Extent
+    import com.cibo.evilplot.plot.aesthetics.DefaultTheme._
+    import com.stripe.rainier.plot._
+
+    val traceplots = Facets(
+      EvilTraceplots.traces(out,
+                            Map("Mu" -> mu, "sigD" -> sigD, "sigE" -> sigE))
+    )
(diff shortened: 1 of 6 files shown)


rainier-plot/src/main/scala/com/stripe/rainier/plot/EvilTraceplots.scala
```diff
@@ -0,0 +1,134 @@
...
+  import com.cibo.evilplot.plot.aesthetics.DefaultTheme._
+  import com.cibo.evilplot.numeric.Point
+
+  /**
+    * Autocorrelation function
+    *
+    * @param x Input data
+    * @param lm Maximum lag required
+    */
+  def acf(x: Seq[Double], lm: Int): List[Double] = {
```
avi-stripe: Is there a way this could share code with https://github.com/stripe/rainier/blob/master/rainier-core/src/main/scala/com/stripe/rainier/sampler/Sampler.scala#L65 ?

rainier-example/src/main/scala/com/stripe/rainier/example/LogReg.scala
```diff
@@ -0,0 +1,100 @@
...
+    // now some EvilPlots
+    import com.cibo.evilplot.plot._
+    import com.cibo.evilplot.geometry.Extent
+    import com.cibo.evilplot.plot.aesthetics.DefaultTheme._
+    import com.stripe.rainier.plot._
+
+    val traceplots = Facets(
+      EvilTraceplots.traces(out, Map("b0" -> beta0, "b1" -> beta1))
+    )
+    javax.imageio.ImageIO.write(
```
avi-stripe: This all feels very boilerplate-y. Could we have a convenience method like `render(plots: List[List[Plot]], path: String, extent: Extent = Extent(1400,1400))`?

rainier-plot/src/main/scala/com/stripe/rainier/plot/EvilTraceplots.scala
```diff
@@ -0,0 +1,134 @@
+/*
+EvilTraceplots.scala
+
+Some diagnostic plots using EvilPlot
+
+ */
+
+package com.stripe.rainier.plot
+
+object EvilTraceplots {
```
avi-stripe: nit: for consistency with `EvilPlot`, I'd prefer `EvilTracePlot`

rainier-example/src/main/scala/com/stripe/rainier/example/Anova.scala
```diff
@@ -0,0 +1,92 @@
...
+
+    // now some EvilPlots
+    import com.cibo.evilplot.plot._
+    import com.cibo.evilplot.geometry.Extent
+    import com.cibo.evilplot.plot.aesthetics.DefaultTheme._
+    import com.stripe.rainier.plot._
+
+    val traceplots = Facets(
+      EvilTraceplots.traces(out,
+                            Map("Mu" -> mu, "sigD" -> sigD, "sigE" -> sigE))
```
avi-stripe: For some reason it took me a while to figure out these were the true values. Can you either assign this to a named val before passing it in, or just use name the param in the method call?

rainier-example/src/main/scala/com/stripe/rainier/example/Anova.scala
```diff
@@ -0,0 +1,92 @@
...
+      current <- prior
+      _ <- RandomVariable.traverse((0 until n) map (addGroup(current, _)))
+    } yield current
+
+    // fit model
+    implicit val rng = ScalaRNG(3)
+    println("Model built. Sampling now (will take a long time)...")
+    val its = 10000
+    val thin = 10000
+    val out = model.sample(HMC(5), 1500000, its * thin, thin)
```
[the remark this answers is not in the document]
darrenjw: Yes!

rainier-plot/src/main/scala/com/stripe/rainier/plot/EvilTraceplots.scala
darrenjw: In principle there ought to be, but it seems like it would involve significantly more code than it would save, so I'm inclined to leave it for now.

rainier-example/src/main/scala/com/stripe/rainier/example/LogReg.scala
darrenjw: Done.

rainier-plot/src/main/scala/com/stripe/rainier/plot/EvilTraceplots.scala
darrenjw: Done.

rainier-example/src/main/scala/com/stripe/rainier/example/Anova.scala
darrenjw: Done.

rainier-example/src/main/scala/com/stripe/rainier/example/LogReg.scala
```diff
@@ -0,0 +1,86 @@
...
+    // now build Rainier model
+    val model = for {
+      beta0 <- Normal(0, 5).param
+      beta1 <- Normal(0, 5).param
+      _ <- Predictor
+        .from { x: Double =>
+          {
+            val theta = beta0 + beta1 * x
+            val p = Real(1.0) / (Real(1.0) + (Real(0.0) - theta).exp)
+            //Bernoulli(p)
```
avi-stripe: Why define the `Bernouilli` and then not use it?

andyscott:
It's a matter of futzing with `crossScalaVersions`. Right now we apply the setting to the whole build. It might be easier to instead apply the setting individually to each project individually, skipping 2.11 for the evil plot project.

avi-stripe:
Ok, thanks, I'll give that a shot.

avi-stripe:
Example plot output for the benefit of bystanders.
pairs
traceplots

rainier-example/src/main/scala/com/stripe/rainier/example/LogReg.scala
darrenjw: Done.

avi-stripe:
@travisbrown-stripe Ok, thanks, let's merge this as is, then, and I look forward to a PR from you later on if you can figure out the better way.

Issue #108, opened 2018-06-01
Sampling still a bit unstable


darrenjw:
I really hate to be *that guy*, but I'm still having some issues with sampling on more challenging models. I've seen CPU usage shoot up, and the sampler also crashes sometimes, without good reason. The example below reliably crashes for me after a couple of minutes. It seems to be an issue deep down in the compute graph.

```scala
import com.stripe.rainier.compute._
import com.stripe.rainier.core._
import com.stripe.rainier.sampler._
import com.stripe.rainier.repl._

object Crash {

  def main(args: Array[String]): Unit = {

    // first simulate some data from a DLM
    val r = new scala.util.Random(0)
    val n = 20 // time points
    val mu = 3.0 // AR(1) mean
    val a = 0.95 // auto-regressive parameter
    val sig = 1.0 // AR(1) SD
    val sigD = 3.0 // observational SD
    val state = Stream
      .iterate(0.0)(x => mu + (x - mu) * a + sig * r.nextGaussian)
      .take(n)
      .toVector
    val obs = state map (_ + sigD * r.nextGaussian)

    // build and fit model
    val prior = for {
      mu <- Normal(5, 10).param
      a <- Normal(1, 0.2).param
      sig <- LogNormal(0, 10).param
      sigD <- LogNormal(0, 10).param
      sp <- Normal(0, 100).param
    } yield Map("mu" -> mu, "a" -> a, "sig" -> sig, "sigD" -> sigD, "SP" -> sp)

    def addTimePoint(current: RandomVariable[Map[String, Real]], i: Int) =
      for {
        map <- current
        os = if (i == 0) map("SP") else map(f"S$i%03d")
        ni = i + 1
        ns <- Normal(map("mu") + map("a") * (os - map("mu")), map("sig")).param
        _ <- Normal(ns, map("sigD")).fit(obs(i))
      } yield map.updated(f"S$ni%03d", ns)

    val fullModel = (0 until n).foldLeft(prior)(addTimePoint(_, _))

    val model = for {
      map <- fullModel
    } yield
      Map(
        "mu" -> map("mu"),
        "a" -> map("a"),
        "sig" -> map("sig"),
        "sigD" -> map("sigD"),
        "SP" -> map("SP")
      )

    implicit val rng = ScalaRNG(4)

    println("Model built. Sampling now...")
    val out = model.sample(HMC(5), 5000, 10000)
    println("Sampling finished.")

  }

}
```

avi-stripe:
[...]

PR #109, opened 2018-06-02
Fix ordering bug and performance issue in Translator

avi-stripe:
This addresses two problems demonstrated by the example code in https://github.com/stripe/rainier/issues/108.

First: when constructing the tree of binary ops for an n-ary sequence of additions or multiplications, we were constructing nodes in a different order from the ultimate traversal - specifically, we were first constructing all of the leaves, then all of the second-last level, and so on up to the root; but traversal will be depth-first, and so (for example), the root's left child will be visited before any of the leaves but constructed after them. In some circumstances this could lead to a the first instance of a particular sub-expression being constructed and made into a variable definition (`VarDef` node), and then a second instance of that same sub-expression being constructed and turned into a reference to that same variable (`VarRef`)... but in fact, when generating code, the `VarRef` will get hit before the `VarDef`. Depending on the situation this could lead to a crash during code gen, or a crash during bytecode verification, or just a subtly incorrect result.

This PR makes a minor change to construct the IRs in the tree lazily by wrapping all of the construction logic in thunks; this results in the actual nodes being created in the same depth-first order as the traversal, which fixes this problem. It also adds a check which will error immediately during the IR construction process if it detects this problem occurring again (though it's not guaranteed to catch every instance of it).

Second, that same example code was extremely slow to compile. The thing that's notable about this model vs others is that because of the long chain auto-regressive chain, some parameters get referenced (transitively/indirectly) over and over and over again, and the Translator was having to rebuild the entire IR tree for these each time (to ultimately find that it was an expression that had been already seen, and so only emit a single VarRef). This PR speeds up that process by caching the Real => IR mapping and directly emitting the reference if we come across the identical Real for a second time. This dramatically speeds up the compilation of this model.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
@@ -175,7 +175,10 @@ private class Translator {
           opt.orElse { cache.get((k, opKey)) }
       }
       hit match {
-        case Some(sym) => VarRef(sym)
+        case Some(sym) =>
+          if (!irKeys.head.collect { case d: VarDef => d }.isEmpty)
+            sys.error("VarRef was used before its VarDef")
+          VarRef(sym)
         case None =>
           val sym = Sym.freshSym()
           cache += (refKeys.head, opKey) -> sym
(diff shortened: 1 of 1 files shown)


darrenjw:
This fixes the bug and improves compilation speed. So the specific example in #108 now runs successfully. My existing models also continue to run fine. So I think this is probably worth merging as-is.

But there are still some issues with the kind of model used in #108 which seem to cause a lot of GC stress and will need further investigation at some point. eg. if you take the example from #108 and bump up the number of time points from `n=20` to `n=50`, then you will see the CPU usage jump massively, along with execution time. It also uses a surprising amount of RAM. I guess it is related to the depth of the compute graph. I wondered if it related to my use of "Map" in the model construction, but I've tried building it without Map, using a case class for the static parameters and a list for the state parameters, and it doesn't seem to help much.

PR #112, opened 2018-06-03
A DLM example

darrenjw:
Example showing how to build and fit a dynamic linear time series state space model.


avi-stripe:
Looks great! Merge at will.

avi-stripe:
@darrenjw out of curiosity, do you have the equivalent to this model written up in Stan or PyMC3? It would be interesting to do a head to head benchmark on something medium-scale like this.

darrenjw:
I don't, but I agree it would be interesting. I often use JAGS for playing around with toy models, but Gibbs samplers are particularly bad for DLMs.

PR #113, opened 2018-06-03
Fix quadratic behavior in `variables` and `gradient`

avi-stripe:
This adds a couple of easy fixes for formerly quadratic behavior that was a problem in compute graphs that referenced the same value a lot of times (like the `DLM` model).

* `Real.variables` keeps track of whether we've visited a node before and bails if we have, rather than happily re-walking that part of the graph umpteen times
* `CompoundDiff` uses a `lazy val` for `toReal` instead of a `def`, so that it still defers computation until the appropriate time but doesn't recompute umpteen times

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
@@ -104,7 +104,15 @@ private object Gradient {
       extends Diff {
     def toReal: Real = {
       val exponent = child.ax(term)
-      gradient.toReal * child * exponent / term
+      val otherTerms =
+        if (child.ax.size == 1)
+          Real.one
+        else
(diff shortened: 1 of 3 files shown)


avi-stripe:
Just to understand: were you seeing successful convergence at n=30 for any previous version of the code? And, similarly, were you seeing success at n=500?

This does likely lead to more code generation than the temporary version which entirely disabled `distribute()`, which is maybe where you were successfully running n=500 before.

Any any rate, I'm aware that I will need to deal with that error case at some point, by splitting into multiple class files. I'll create an issue for it.

avi-stripe:
Thanks, that's helpful.

avi-stripe:
That would have been this commit BTW: https://github.com/stripe/rainier/pull/113/commits/c11188891009b0131747d81158afbdd078c8e9ae

PR #119, opened 2018-06-04
tweak DLM to be more numerically stable

avi-stripe:
The issues we happened to see on master with DLM at n=30 were, I'm pretty sure, just a coincidental symptom of an overall numerical instability of DLM's gradient; it was very sensitive to the particular ordering of terms (which were ultimately at the whim of the JVM). This slight adjustment seems to make it better.

I'm pretty certain we can automatically detect this kind of problem and flag it. I don't understand it well enough yet to know if there's a practical way for us to automatically avoid it entirely.

### DIFF
--- a/rainier-example/src/main/scala/com/stripe/rainier/example/DLM.scala
+++ b/rainier-example/src/main/scala/com/stripe/rainier/example/DLM.scala
@@ -49,7 +49,8 @@ object DLM {
         static = tup._1
         states = tup._2
         os = states.head
-        ns <- Normal(static.mu + static.a * (os - static.mu), static.sig).param
+        ns <- Normal(((Real.one - static.a) * static.mu) + (static.a * os),
+                     static.sig).param
         _ <- Normal(ns, static.sigD).fit(obs(i))
       } yield (static, ns :: states)
 


avi-stripe:
Also, the traces look better now. For comparison, traceplots from when it was "working" before:
traceplots
Vs when it was definitely not working:
traceplots
Vs now:
traceplots

avi-stripe:
Yeah, agreed that it's unnerving. To be clear, every time you run `sample` it is in fact re-compiling, and this is where the ordering of objects inside a `Map` can introduce non-determinism that is unrelated to the `RNG`. If there were an API for compiling the sampler once and then using it multiple times (which, as a side note, would save some time when we do multiple parallel chains like in `sampleWithDiagnostics`), then you would see identical traces.

Getting rid of that non-determinism entirely would certainly be possible, though not easy. How valuable do you think that would be? If we did that I'd want to optionally introduce a way to do a targeted randomization to detect exactly this kind of problem.

darrenjw:
It's something that would be nice, but I'm not sure it's vital. In general I'm not a big fan of setting random number seeds - I don't want my results to be a fluke of the seed I happened to pick. But I guess that a lack of determinism could cause some issues with testing at some point.

Issue #115, opened 2018-06-04
Add benchmarks for DLM


avi-stripe:
The newly-added `DLM` model has been good at shaking out various performance issues due to its deep compute graph. We should add it to the compilation and execution benchmarks.

PR #124, opened 2018-06-06
standardize compute benchmarks

avi-stripe:
The reworks the benchmarks to be a little easier to work with, so that adding a new expression just means defining a subclass that implements `expression`, and all the benchmarks for various tasks like building, compiling, running, and deriving the gradient of that expression will all automatically get added.

It also dumps a `benchmarks.txt` into the repo with the output of a run; the idea is that performance-affecting changes will update this file so that we get to see the diffs in the PRs.


avi-stripe:
Oh and per https://github.com/stripe/rainier/issues/115 it adds a benchmark for `DLM`, which is by far the most complex computation in the benchmark suite now.

PR #125, opened 2018-06-06
Limit `pow` to integer exponents

avi-stripe:
This makes one tiny fix and one fairly major change/fix to the compute graph.

The tiny one is that in rare cases we were letting through some 0 coefficients in `Line` during `distribute`, which then wreak havoc later on. We now make sure to use `merge` to avoid this.

The major one is that we now represent exponents in a `LogLine` as integers. This avoids problems with floating point precision as we aggregate or factor out exponent terms - you really don't want (as I was able to trigger) to end up with `x^0.99999999` instead of `x`, especially if `x` can be negative. It does mean that we can't take square roots, which I suspect we will one day want to do, but it would be easy to add a unary `SqrtOp` whenever we need it.

### DIFF
--- a/rainier-core/src/test/scala/com/stripe/rainier/compute/RealTest.scala
+++ b/rainier-core/src/test/scala/com/stripe/rainier/compute/RealTest.scala
@@ -7,24 +7,30 @@ class RealTest extends FunSuite {
   def run(description: String)(fn: Real => Real): Unit = {
     test(description) {
       val x = new Variable
+      val x2 = new Variable
       val result = fn(x)
+      val result2 = fn(x2)
       val c = Compiler.default.compile(List(x), result)
       List(1.0, 0.0, -1.0, 2.0, -2.0, 0.5, -0.5).foreach { n =>
         val constant = fn(Constant(n))
-        assert(constant.isInstanceOf[Constant], s"[n=$n]")
-        val eval = new Evaluator(Map(x -> n))
+        assert(constant.isInstanceOf[Constant], s"[c, n=$n]")
+        val eval = new Evaluator(Map(x -> n, x2 -> n))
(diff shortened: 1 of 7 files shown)


avi-stripe:
The alternative would be to use `BigDecimal` everywhere inside `Real` to keep everything precise during partial evaluation (still using double in the compiled code for performance).

rainier-core/src/test/scala/com/stripe/rainier/compute/RealTest.scala
```diff
@@ -7,24 +7,30 @@ class RealTest extends FunSuite {
...
         val withVar = eval.toDouble(result)
         assertWithinEpsilon(constant.asInstanceOf[Constant].value,
                             withVar,
-                            s"[n=$n]")
+                            s"[c/ev, n=$n]")
         val compiled = c(Array(n))
-        assertWithinEpsilon(withVar, compiled, s"[ir, n=$n]")
+        assertWithinEpsilon(withVar, compiled, s"[ev/ir, n=$n]")
+        val withVar2 = eval.toDouble(result2)
+        assertWithinEpsilon(withVar, withVar2, s"[ev/ev, n=$n]")
```
roban-stripe: OOC, what failure mode is this test getting at?

avi-stripe: This is actually sorta aspirational in that I don't yet have any sufficiently simple test cases that I can get this to fail with. But the idea was to test numerical instability in results from random reordering of terms when building the DAG.

avi-stripe:
After benchmarking the `BigDecimal` approach, the cost seems low enough that I think it's worth it, since it also has some other benefits. So I'm closing this one in favor of a WIP BigDecimal PR.

PR #129, opened 2018-06-07
`fit` returns a `Likelihood` rather than a `Generator`

avi-stripe:
This is theoretically a breaking change, though I suspect very little if any code in the wild will be affected.

Previously, `fit` returned a `RandomVariable[Generator[T]]`, with the idea being that it would be a generator of synthetic data from the posterior that was equivalent, as closely as possible, to the observations passed into `fit`.

I originally made this choice because it seemed like the only useful thing that could be *generically* returned from `fit`; my reasoning was that all you could do with a generic `Likelihood[T]` was call `fit` on it again, which you have no reason to do, whereas even if that generator wasn't always useful (or even usually useful) it was better than nothing.

There were two problems with this reasoning:
* if you have large number of observations, it turns out it can be extremely expensive to create the `Generator`, which is really lame if you're just going to throw it away (which you almost certainly are going to do in that specific case)
* Scala's type system is smarter than I gave it credit for, and if `Distribution[T]` and `Predictor[T]` provide narrower return types in their implementation of `fit`, then those will be respected at the call site. This means that it *is* useful to return `RandomVariable(this, likelihood)` from `fit`, because you have can have a less-generic type attached to `this`, and can do useful things with it (like call `generator` in the `Distribution` case or call `predict` in the `Predictor` case).

This change includes a significant update to `tour.md` to explain the new API. As part of doing this, I decided to switch the default `Sampler` from `Walkers` to `HMC`, since I was getting better plots in the tour output by doing that.

### DIFF
--- a/docs/tour.md
+++ b/docs/tour.md
@@ -18,7 +18,7 @@ val model = for {
     slope <- LogNormal(0,1).param
     intercept <- LogNormal(0,1).param
     regression <- Predictor.from{x: Int => Poisson(x*slope + intercept)}.fit(data)
-} yield regression
+} yield regression.predict(21)
 ```
 
 ## Distribution
@@ -33,7 +33,7 @@ You construct a distribution from its parameters, usually with the `apply` metho
 
 ```scala
 scala> val normal: Distribution[Double] = Normal(0,1)
-normal: com.stripe.rainier.core.Distribution[Double] = com.stripe.rainier.core.Injection$$anon$1@4614111e
+normal: com.stripe.rainier.core.Distribution[Double] = com.stripe.rainier.core.Injection$$anon$1@4ce0cc1a
 ```
 
 In Rainier, `Distribution` objects play three different roles. `Continuous` distributions (like `Normal`), implement `param`, and all distributions implement `fit` and `generator`. Each of these methods is central to one of the three stages of building a model in Rainier: defining your parameters and their priors; fitting the parameters to some observed data; and using the fit parameters to generate samples of some posterior distribution of interest. We'll start by exploring each of these in turn.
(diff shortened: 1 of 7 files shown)


roban-stripe:
Nice that this involves deleting more code than it adds.

docs/tour.md
```diff
@@ -230,135 +231,149 @@ scala> val lr = (poisson9.density - poisson10.density).exp
...
+        4.68     5.55     6.42     7.28     8.15     9.02     9.88    10.75    11.61  
+```
+
+Looks like our daily rate is probably somewhere between 6 and 9, which corresponds well to the steep 3x dropoff we saw before between 9 and 10.
+
+## A mathematical digression
+
+Please feel free to skip this section if it's not helping you.
+
+One thing that sometimes confuses people at this stage is, what is the line with `fit` actually doing? It seems to return a `poisson` object that we just ignore. (That object, by the way, is just the `Poisson(r)` distribution object that we used to do the `fit`.) And yet the line is clearly doing something, because sampling from `rate` gives quite different results from just sampling directly from `e_x`.
```
stephanie-stripe: thanks for adding this clarifying section!

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -42,9 +42,20 @@ class RandomVariable[+T](val value: T,
...
   def sample[V]()(implicit rng: RNG, sampleable: Sampleable[T, V]): List[V] =
-    sample(Sampler.Default.sampler,
-           Sampler.Default.warmupIterations,
-           Sampler.Default.iterations)
+    sample(Sampler.Default.iterations)
+
+  def sample[V](iterations: Int)(implicit rng: RNG,
+                                 sampleable: Sampleable[T, V]): List[V] =
+    if (density.variables.isEmpty) {
+      val fn = sampleable.prepare(value, density.variables)
```
stephanie-stripe: hmm, would it make sense to avoid the warmup within the more generic version of `sample` below? then this one can just call `sample(...)` with all the defaults.

PR #130, opened 2018-06-07
allow for multiple class files

avi-stripe:
Instead of using a single class for a single `CompiledFunction`, we now generate any number of helper classes that have static methods the main class can call into; this avoids the "class file too large" errors.


mio-stripe:
LGTM

PR #131, opened 2018-06-08
More stable benchmark output

avi-stripe:
This just adds a formatting script to post-process the raw `jmh` output into something that should be more stable (in particular, by rounding the numbers to have sig figs in keeping with the reported variance). The hope is that this will make diffs smaller between runs when nothing has materially changed.


rainier-benchmark/format.rb
```diff
@@ -0,0 +1,36 @@
...
+    @klass = method_parts[0].split("Benchmark")[0]
+    @method = method_parts[1]
+    @timing = parts[4].to_f
+    @stddev = parts[6].to_f
+  end
+
+  attr_reader :klass, :method
+
+  def rounded_time
+    oom = 10 ** (Math.log(@stddev) / Math.log(10)).to_i
```
roban-stripe: Isn't there a `Math.log10`?

avi-stripe: TIL! Will fix.

PR #132, opened 2018-06-08
Add cats and scalacheck modules

andyscott:
Adds a cats module and a scalacheck module with some very basic and possibly incorrect instances.

Consider this WIP/partially baked. Feedback welcomed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,143 @@
+lazy val root = project.
+  in(file(".")).
+  aggregate(rainierCore, rainierPlot, rainierCats, rainierScalacheck).
+  aggregate(rainierDocs, rainierExample).
+  aggregate(rainierBenchmark, rainierTests).
+  aggregate(shadedAsm).
+  settings(unpublished)
+
+/* dependency versions */
+lazy val V = new {
+  val asm = "6.0"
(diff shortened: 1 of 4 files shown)


avi-stripe:
Nice! This looks great.

One thing that would be really nice for the scalacheck module is a `Gen[Real=>Real]`. This should be very similar to the `Gen[Real]` and in fact I think you could trivially build `Gen[Real]` on top of it. There's a couple of uses I have in mind for this:

* It helps with a `Gen[RandomVariable]` since you can plug the `Real=>Real` into `map` or build a fn for `flatMap` from it
* Given an arbitrary `fn: Real=>Real`, `v: Double`, and `x: Variable`, it's extremely valuable to be able to check the equivalence of `fn(Real(v))` and `Evaluator(Map(x->v)).toDouble(fn(x))`

andyscott:
That should be easy to add. I can also add the test for `Evaluator(Map(x->v)).toDouble(fn(x))`.

andyscott:
I have some separate code for testing the Evaluator. It's still a work in progress, since generating real expressions is a good bit of work.

avi-stripe:
LGTM

PR #134, opened 2018-06-08
Use BigDecimal inside Real

avi-stripe:
This changes all of the internal constants in the `Real` DAG to use `BigDecimal` instead of `Double`. This means that all of the partial evaluation/constant folding, and anything using `Evaluator`, will use higher precision arithmetic. Although it will be a bit slower, this was already expected to be the slow path, so that seems ok. Anything using `Compiler` will still use `double` and will run at the same speed as before.


avi-stripe:
This is ready for review now, just waiting to update the benchmarks.

PR #140, opened 2018-06-10
Add the Beta distribution.

cqfd:
Here's the Beta distribution!

I've also added a tweak to `Combinatrics.gamma` suggested by Boost's lgamma function: https://www.boost.org/doc/libs/1_50_0/libs/math/doc/sf_and_dist/html/math_toolkit/special/sf_gamma/lgamma.html. Our existing approximation to the log gamma function isn't very accurate when z is small, but amusingly, there's an easy fix: just do lgamma(z + 1) - log(z)!

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -147,7 +147,7 @@ object Exponential {
 object Beta {
   def apply(a: Real, b: Real): Continuous = new Continuous {
     def realLogDensity(real: Real): Real =
-      If(real < 0, Real.negInfinity, If(real > 1, Real.negInfinity, betaDensity(real)))
+      Uniform.standard.realLogDensity(real) + betaDensity(real)
 
     def param: RandomVariable[Real] =
       Uniform.standard.param.flatMap { u =>
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -161,7 +161,7 @@ object Beta {
       RandomVariable(logistic, density)
     }
 
-    def generator: Generator[Double] =
+    val generator: Generator[Double] =
       Gamma(a, 1).generator.zip(Gamma(b, 1).generator).map {
         case (z1, z2) =>
           z1 / (z1 + z2)
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -161,11 +161,11 @@ object Beta {
       RandomVariable(logistic, density)
     }
 
-    def generator: Generator[Double] = Generator.require(Set(a, b)) { (r, n) =>
-      val z1 = Gamma(a, 1).generator.get(r, n)
-      val z2 = Gamma(b, 1).generator.get(r, n)
-      z1 / (z1 + z2)
-    }
+    def generator: Generator[Double] =
+      Gamma(a, 1).generator.zip(Gamma(b, 1).generator).map {
+        case (z1, z2) =>
+          z1 / (z1 + z2)
+      }
   }
 }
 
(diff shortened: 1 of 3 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -141,6 +141,34 @@ object Exponential {
...
+object Beta {
+  def apply(a: Real, b: Real): Continuous = new Continuous {
+    def realLogDensity(p: Real): Real =
+      If(p < 0,
+         Real.zero.log,
+         If(p > 1,
+            Real.zero.log,
+            (a - 1) * p.log + (b - 1) * (1 - p).log - Combinatrics.beta(a, b)))
+
+    def param: RandomVariable[Real] = {
```
avi-stripe: The duplication with `Uniform` for `realLogDensity` and `param` bothers me a bit. Two possible solutions:

* define `Uniform.standard` as `Beta(1,1)`. If you did this, you'd be making uniform's logDensity slightly different from 0.0 because of the approximate gamma function; not sure if this ever matters in practice. Also, you'd probably want to special case the Beta.generator for the (1,1) case to just use `r.standardUniform` (but maybe you want to do that anyway). But there's a certain elegance to it.
* or, do something like defining Beta's `realLogDensity` as `Uniform.standard.logDensity(p) + betaDensity(p)`, and define `param` as `Uniform.standard.flatMap{u => RandomVariable(u, betaDensity(u))}`. Then put the ` `(a - 1) * p.log + (b - 1) * (1 - p).log - Combinatrics.beta(a, b)))` inside a private `betaDensity` method. This is maybe less elegant but more practical.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -141,6 +141,34 @@ object Exponential {
...
+      val x = new Variable
+      val logistic = Real.one / (Real.one + (x * -1).exp)
+      val logisticJacobian = logistic * (1 - logistic)
+      val density = realLogDensity(logistic) + logisticJacobian.log
+      RandomVariable(logistic, density)
+    }
+
+    def generator: Generator[Double] = Generator.require(Set(a, b)) { (r, n) =>
+      val z1 = Gamma(a, 1).generator.get(r, n)
```
avi-stripe: it's slightly wasteful to create `Gamma(a,1).generator` (and the one for b) for each sample. You can pull these out of the closure and into temps in the `generator` method body.

avi-stripe: In fact, you can do this more cleanly with a `zip` and `map` on those two generator objects.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -21,6 +21,12 @@ trait Generator[T] { self =>
     def get(implicit r: RNG, n: Numeric[Real]): U = fn(self.get).get
   }
 
+  def zip[U](other: Generator[U]): Generator[(T, U)] =
+    for {
```
avi-stripe: This definition works but it loses track of `other.requirements`. It would be better to keep them.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -146,26 +146,22 @@ object Exponential {
...
 object Beta {
   def apply(a: Real, b: Real): Continuous = new Continuous {
-    def realLogDensity(p: Real): Real =
-      If(p < 0,
-         Real.negInfinity,
-         If(p > 1,
-            Real.negInfinity,
-            (a - 1) * p.log + (b - 1) * (1 - p).log - Combinatrics.beta(a, b)))
+    def realLogDensity(real: Real): Real =
+      If(real < 0, Real.negInfinity, If(real > 1, Real.negInfinity, betaDensity(real)))
```
avi-stripe: I think this could just be `Uniform.standard.realLogDensity(real) + betaDensity(real)`.

PR #144, opened 2018-06-11
Some exact gamma cases, etc

avi-stripe:
cc @cqfd 

This is a bit of a grab bag of related numerical things.

* For gamma(0), gamma(1), and gamma(2), we return the exact values instead of the approximation. (Currently we just do this statically; we could also decide to make these checks dynamically, at a small performance cost).
* This allows us to use Beta(1,1) from Uniform without loss of performance or precision, which is satisfying. (We still special case the `generator` though).
* In the process if doing this, I noticed some other special cases we needed after the BigDecimal change; most notably, `Real.zero.log` would crash instead of returning `NegInfinity`. These are fixed.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
@@ -30,7 +30,7 @@ private[compute] object RealOps {
                 "Cannot take the log of " + value.toDouble)
             else
               Real(Math.log(value.toDouble))
-          case AbsOp => Real(Math.abs(value.toDouble))
+          case AbsOp => Real(value.abs)
         }
       case nc: NonConstant =>
         val opt = (op, nc) match {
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
```diff
@@ -6,12 +6,30 @@ private[compute] object RealOps {
...
         op match {
           case ExpOp => Real(Math.exp(value.toDouble))
-          case LogOp => Real(Math.log(value.toDouble))
+          case LogOp =>
+            if (value.toDouble < 0)
+              throw new ArithmeticException(
+                "Cannot take the log of " + value.toDouble)
+            else
+              Real(Math.log(value.toDouble))
           case AbsOp => Real(Math.abs(value.toDouble))
```
cqfd: I'm assuming we have to convert to a double for `LogOp` since seems like there's no built-in way to take the log of a BigDecimal? For `AbsOp` though, could this be `Constant(value.abs)`?

avi-stripe: Right - nice catch, thank you.

PR #145, opened 2018-06-12
Jglidden gkk variational inference

jglidden-stripe:
cc @gkk-stripe


avi-stripe:
Can you rebase this against master?

rainier-core/src/main/scala/rainier/compute/Real.scala
```diff
@@ -69,6 +69,21 @@ object Real {
     loop(real, Set.empty)
   }
 
+  def substituteVariable(x: Real, f: Variable => Real): Real = x match {
```
avi-stripe: This is great. You should memoize the transformation to avoid wasted work and so that nodes that start out identical end up identical.

avi-stripe: What's the reason for taking `Variable => Real` instead of `Map[Variable,Real]` which is what you are actually providing - just that the former seems more general? I think I'd lean somewhat towards the concreteness/explicitness of the latter.

rainier-core/src/main/scala/rainier/sampler/Variational.scala
```diff
@@ -0,0 +1,127 @@
...
+case class MeanFieldNormalGuide(targetVariables: Seq[Variable])(
+    implicit rng: RNG)
+    extends Guide {
+  private val normals: Map[(Variable, Variable), (Real, Continuous)] =
+    targetVariables
+      .map(tv => {
+        val mu = new Variable
+        val sigma = new Variable
+        val dist = Normal(mu, sigma)
+        ((mu, sigma), (dist.param.density.variables.head, dist))
```
avi-stripe: tiny style nit but when constructing maps I like to use the `->` tuple constructor to make key/value pairs clear

rainier-core/src/main/scala/rainier/sampler/Variational.scala
```diff
@@ -0,0 +1,127 @@
...
+      })
+      .toMap
+
+  override def params: Seq[Variable] = normals.keys.toSeq.flatMap {
```
avi-stripe: you don't need `override` when the trait just declared the signature without a method body

rainier-core/src/main/scala/rainier/sampler/Variational.scala
```diff
@@ -0,0 +1,127 @@
...
+
+case class MeanFieldNormalGuide(targetVariables: Seq[Variable])(
+    implicit rng: RNG)
+    extends Guide {
+  private val normals: Map[(Variable, Variable), (Real, Continuous)] =
+    targetVariables
+      .map(tv => {
+        val mu = new Variable
+        val sigma = new Variable
+        val dist = Normal(mu, sigma)
```
avi-stripe: the way you deconstruct `Normal` here to reach in and grab its variable and later to use `realLogDensity` which is meant to be private feels to me like you shouldn't really be using `Normal` at all and should just be redefining the normal pdf locally to this class - it's not that complex a function.

rainier-core/src/main/scala/rainier/sampler/Variational.scala
```diff
@@ -0,0 +1,127 @@
...
+  override def sample(
+      paramSettings: Map[Variable, Double]): Map[Variable, Double] = {
+    (targetVariables zip (generate zip normals).map {
+      case (eps, ((mu, sigma), _)) => {
+        paramSettings(mu) + paramSettings(sigma) * eps
+      }
+    }).toMap
+  }
+  override def init: Map[Variable, Double] = {
+    normals.flatMap {
```
avi-stripe: wouldn't this be simpler just as `params.map{v => v -> 0.0}.toMap` ?

rainier-core/src/main/scala/rainier/sampler/Variational.scala
```diff
@@ -0,0 +1,127 @@
...
+        mu + sigma * eps
+      }
+    })).toMap
+  }
+  override def generate(implicit rng: RNG) = {
+    normals.map(_ => rng.standardNormal).toSeq
+  }
+  override def sample(
+      paramSettings: Map[Variable, Double]): Map[Variable, Double] = {
+    (targetVariables zip (generate zip normals).map {
```
avi-stripe: style nit: please use dots for method calls

avi-stripe (commented):
Haven't fully processed everything here but left some initial comments. Very cool to have a start at this.

avibryant:
I'm going to close this because it's so out of date relative to the current develop branch, but I'd be excited to see more progress made on it some day.

PR #146, opened 2018-06-12
Beta-binomial

avi-stripe:
Adds the beta-binomial distribution.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -144,31 +144,33 @@ object Exponential {
 /**
   * A Beta distribution with expectation `a/(a + b)` and variance `ab/((a + b)^2 (1 + a + b))`.
   */
-object Beta {
-  def apply(a: Real, b: Real): Continuous = new Continuous {
-    def realLogDensity(real: Real): Real =
-      If(real >= 0,
-         If(real <= 1, betaDensity(real), Real.negInfinity),
-         Real.negInfinity)
+final case class Beta(a: Real, b: Real) {
+  def realLogDensity(real: Real): Real =
+    If(real >= 0,
+       If(real <= 1, betaDensity(real), Real.negInfinity),
+       Real.negInfinity)
 
-    def param: RandomVariable[Real] = {
-      val x = new Variable
-      val logistic = Real.one / (Real.one + (x * -1).exp)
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -144,34 +144,40 @@ object Exponential {
...
-      (a - 1) *
-        u.log + (b - 1) *
-        (1 - u).log - Combinatorics.beta(a, b)
+  def binomial: Predictor[Int, Int, BetaBinomial] = Predictor.from { k: Int =>
+    BetaBinomial(a, b, k)
   }
 }
 
+object Beta {
+  def from(mean: Real, precision: Real) = Beta(mean * precision, (Real.one - mean) * precision)
```
avi-stripe: I'd love a more descriptive name for this method but `Beta.meanAndPrecision` seems too long and I can't think of anything better.

roban-stripe: `fromMoments` maybe (though I guess precision is the inverse of the second moment). or `fromSummaryStats`, `fromStats`?

avi-stripe: Hm, should we make it (mean,variance) instead and call it `Beta.moments` then?

roban-stripe: yeah, that's be fine with me

roban-stripe:
LGTM, aside from any renaming you want to do on `Beta.from`.

roban-stripe:
haha, why not both? great!

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -175,7 +175,9 @@ final case class Beta(a: Real, b: Real) {
 }
 
 object Beta {
-  def from(mean: Real, precision: Real) = Beta(mean * precision, (Real.one - mean) * precision)
+  def meanAndPrecision(mean: Real, precision: Real) = Beta(mean * precision, (Real.one - mean) * precision)
+  def meanAndVariance(mean: Real, variance: Real) =
+    meanAndPrecision(mean, mean * (Real.one - mean) / variance - 1)
```
roban-stripe: Wait, @avi-stripe, isn't precision just 1/variance?

PR #148, opened 2018-06-13
introduce compute.Context

avi-stripe:
This is a small refactoring setting things up for some more consequential changes.

This introduces a `Context` object which wraps a `density` (or, more generally, some "output" `Real`), and provides a single bottleneck for performing static analysis on it - currently that means obtaining the `variables` and `gradient` for that real, and a `compiler` to compile it or related functions with.

For now this just delegates to the same implementations as before and doesn't provide any actual value. However, in the future I intend to use this as a place to more efficiently manage the various caches and lookup tables used in the analysis and compilation process.


mio-stripe:
Nice
LGTM

PR #154, opened 2018-06-20
Make Poisson generate non-negative numbers

mio-stripe:
Previously, a `Poisson(0.0).generator` would return `-1`s. Now it returns `0`s.


avi-stripe:
LGTM

PR #155, opened 2018-06-20
bug fixed

mikeheaton-stripe:
Fixes a bug in the Multinomial pdf, which would cause an error when an outcome had `p=0` and 0 outcomes.

Also adds a test.


rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -85,7 +86,13 @@ final case class Multinomial[T](pmf: Map[T, Real], k: Int)
   def logDensity(t: Map[T, Int]): Real =
     Combinatorics.factorial(k) + Real.sum(t.toList.map {
       case (v, i) =>
-        i * pmf.getOrElse(v, Real.zero).log - Combinatorics.factorial(i)
+        val p = pmf.getOrElse(v, Real.zero)
+        val pTerm = if (i == 0 & p == Real.zero) {
```
avi-stripe: This only works if `p` is the constant 0. If it's unknown, you can still end up with `p.log` being evaluated to -inf at runtime. You need to use an `If()` expression instead.

mikeheaton-stripe: Gottit. How's this?

avi-stripe:
LGTM!

PR #157, opened 2018-06-21
Remove unused tmp file generation and associated dependency

andyscott:


mio-stripe:
I wonder if these temp files are useful for debugging? Easy enough to revert or add back if we want them later.

andyscott:
They might be useful for debugging. If they are added back then we should:
- use java's tmp file path creation
- also avoid adding an additional dependency! especially just for debugging purposes

Issue #159, opened 2018-06-21
Wrong sign of certain numbers


KonstantinBob:
At first, I would like to thank you for your work as it is extremely useful to me.

Now for the issue:
Running the following commands from the root directory of the current repository
```scala
sbt
project rainierPlot
console
import com.stripe.rainier.compute._
(5d to 6d by 0.1d).foreach(x=>if(Real(x).<=(0)==Real.zero){println(x+" >= 0")}else{println(x+" < 0 !------------!")})
```
yields the following output:
```
5.0 >= 0
5.1 >= 0
5.199999999999999 >= 0
5.299999999999999 >= 0
5.399999999999999 >= 0
5.499999999999998 < 0 !------------!
5.599999999999998 >= 0
5.6999999999999975 >= 0
5.799999999999997 >= 0
5.899999999999997 >= 0
5.9999999999999964 >= 0
```
This is just a minimal working example, the effect occurred also in other circumstances.

avi-stripe:
Thanks so much for the bug report! I'll prepare a fix later today.

In case anyone's curious about what's going on, you can see evidence of the problem just with this:

```
scala>  Real(5.499999999999998)  /  Real(5.499999999999998) 
res10: com.stripe.rainier.compute.Real = Constant(0.9999999999999999999999999999999998)
```

Note that this only happens with constants, not with `Evaluator` or `Compiler`:

```
scala> (new Evaluator(Map(x -> 5.499999999999998))).toDouble(x/x)
res11: Double = 1.0
```

Ultimately this is because, due to a quirk of the way `Real` is implemented, for a constant x, `x/y` is computed as `x.pow(-1) * y`. And:

```
scala> BigDecimal(5.499999999999998).pow(-1) * BigDecimal(5.499999999999998)
res12: scala.math.BigDecimal = 0.9999999999999999999999999999999998
```

KonstantinBob:
Thank you for your quick response.
Interesting to see where the problem was rooted.
Best regards and good luck with your further work on rainier, I am excited to see what the future brings.

PR #160, opened 2018-06-21
use actual / on BigDecimal for constant divisions

avi-stripe:
This adds a test for #159 as well as a simple fix.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
@@ -90,6 +90,12 @@ private[compute] object RealOps {
         LogLineOps.multiply(LogLine(nc1), LogLine(nc2))
     }
 
+  def divide(left: Real, right: Real): Real =
+    (left, right) match {
+      case (Constant(x), Constant(y)) => Real(x / y)
+      case _                          => left * right.pow(-1)
+    }
+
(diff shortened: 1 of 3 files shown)


mio-stripe:
LGTM

rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
```diff
@@ -90,6 +90,12 @@ private[compute] object RealOps {
         LogLineOps.multiply(LogLine(nc1), LogLine(nc2))
     }
 
+  def divide(left: Real, right: Real): Real =
+    (left, right) match {
+      case (Constant(x), Constant(y)) => Real(x / y)
```
mio-stripe: Actually, in the repl, `1.0 / 0.0` gives `Infinity` whereas `BigDecimal(1.0) / BigDecimal(0.0)` throws an exception. We should guard against this.

mio-stripe: Yep, `Real(1.0) / Real(0.0)` throws an exception.

avi-stripe: Addressed in the most recent commit, can you take another look @mio-stripe ?

mio-stripe:
LGTM

PR #161, opened 2018-06-21
Optimizer trait

avi-stripe:
This lays the low-level groundwork for mini-batch-based optimization rather than full sampling, including a very simple SGD example. Primarily, this includes:

- tracking "placeholder" variables (vs parameter variables) in `RandomVariable` along with columnar data for each one.
- ensuring that we don't bother with gradients on placeholder vars
- an `optimize` method on `RandomVariable` and an `Optimize` trait with a simple `SGD` implementation that can use those placeholder variables and columns of data to do mini-batch optimization

However, there is not yet any high-level API comparable to `Distribution.fit` for producing RVs that have placeholders, or density functions that make use of those placeholders. This is manually hacked together inside the `example.SGD` which is currently the only thing that exercises this code.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
@@ -2,7 +2,8 @@ package com.stripe.rainier.compute
 
 case class Context(density: Real, placeholders: Seq[Variable]) {
   val compiler: Compiler = IRCompiler(200, 100, false)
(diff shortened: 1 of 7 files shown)


avi-stripe:
The (possibly better) alternative to this is if I can find a reasonable way to express a single level of vectorization without otherwise damaging the API too much.

avi-stripe:
Pursuing a slightly different tack.

Issue #135, opened 2018-06-09
add `record` and `replay` to `RandomVariable`


avi-stripe:
The idea being that instead of immediately using a bunch of sampled parameter vectors, you can store them and use them later to generate samples from the same or a compatible RV with redoing any inference.

PR #162, opened 2018-06-22
[WIP] Record and replay models

mio-stripe:
Add serialization and deserialization for models.
Addresses https://github.com/stripe/rainier/issues/135

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Show.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Show.scala
@@ -1,0 +1,41 @@
+package com.stripe.rainier.compute
+
+/**
+  * Trait for serializable and deserializable types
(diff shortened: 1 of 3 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Show.scala
```diff
@@ -0,0 +1,38 @@
+package com.stripe.rainier.compute
+
+/**2
+  * Trait for serializable and deserializable types
+  */
+trait Show[-T, S] {
```
andyscott: This name is commonly used for printing to String, so using it for serialization/deserialization seems confusing IMHO.

mio-stripe:
Going to scope this down, so closing this. I'll keep comments in mind if I revive in the future.

PR #163, opened 2018-06-26
Uncross and remove Scala dep from shaded ASM jar

andyscott:
This does a few things:
- Converts the shaded ASM module and drops the Scala version suffix. This is a wrapper project for a Java dependency, so the suffix isn't needed.
- Removes the Scala library as a dependency for shaded ASM.
- Versions shaded ASM separately from Rainier. ASM versions will change automatically when we bump the underlying ASM version, or if we manually edit the suffix. We should manually edit the suffix if we change the assembly settings.
- Cleans up how Scala cross versions are handled and makes sure tests are run for both versions of Scala.

### DIFF
--- a/.rainier-shaded-asm/version.sbt
+++ b/.rainier-shaded-asm/version.sbt
@@ -1,1 +1,0 @@
-version in ThisBuild := "0.1.2-SNAPSHOT"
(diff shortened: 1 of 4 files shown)


avi-stripe:
Looks fine but can you say something about what this is doing?

andyscott:
Notes added now that this is ready for review (assuming CI passes this time 'round).

PR #165, opened 2018-06-27
WIP: Mixture distributions

mikeheaton-stripe:
WIP: adds support for continuous Mixture Distributions. 

This requires factoring out the support transform and the logJacobian of the support transform for each Continuous distribution, so they can be exposed to the mixture distribution class. As a bonus, we get to make the `param` method into a pattern.

APIwise, the Mixture needs to know what support transform (and its Jacobian) to use. I implemented creating from a `Map[Continuous, Real]` if the transform is specified explicitly, or from a `Seq[(Continuous, Real)]` if not; in this case the support of the first distribution is used.

This passes the existing tests, but I haven't written tests specifically for these changes yet.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -8,8 +8,6 @@ import scala.annotation.tailrec
   * A Continuous Distribution, with method `param` allowing conversion to a RandomVariable
   */
 trait Continuous extends Distribution[Double] { self =>
-  def param: RandomVariable[Real]
-
   def logDensity(t: Double): Real =
     realLogDensity(Real(t))
 
@@ -18,6 +16,20 @@ trait Continuous extends Distribution[Double] { self =>
   def exp: Continuous = Exp.transform(this)
 
   private[rainier] def realLogDensity(real: Real): Real
+
+  private[rainier] def supportTransform(v: Variable): Real
+  private[rainier] def supportTransformLogJacobian(v: Variable): Real
+
+  def param: RandomVariable[Real] = {
+    val x = new Variable
+
+    val support = supportTransform(x)
+    val logJacobian = supportTransformLogJacobian(x)
+    val logDensity = logJacobian + realLogDensity(support)
+
+    RandomVariable(support, logDensity)
+  }
+
 }
 
 /**
@@ -34,10 +46,9 @@ trait LocationScaleFamily { self =>
       }
     def realLogDensity(real: Real): Real =
       self.logDensity(real)
-    def param: RandomVariable[Variable] = {
-      val x = new Variable
-      RandomVariable(x, self.logDensity(x))
-    }
+
+    def supportTransform(v: Variable): Real = v
+    override def supportTransformLogJacobian(v: Variable): Real = Real.zero
   }
 
   def apply(location: Real, scale: Real): Continuous =
@@ -94,10 +105,8 @@ object Gamma {
     This is pdf(f(x))f'(x) which is pdf(e^x)e^x.
     If we take the logs we get logPDF(e^x) + x.
      */
-    def param: RandomVariable[Real] = {
-      val x = new Variable
-      RandomVariable(x.exp, x + realLogDensity(x.exp))
-    }
+    override def supportTransform(v: Variable): Real = v.exp
+    override def supportTransformLogJacobian(v: Variable): Real = v
 
     def generator: Generator[Double] = Generator.require(Set(shape)) { (r, n) =>
       val a = n.toDouble(shape)
@@ -150,13 +159,10 @@ final case class Beta(a: Real, b: Real) extends Continuous {
        If(real <= 1, betaDensity(real), Real.negInfinity),
        Real.negInfinity)
 
-  def param: RandomVariable[Real] = {
-    val x = new Variable
-    val logistic = Real.one / (Real.one + (x * -1).exp)
-    val logisticJacobian = logistic * (1 - logistic)
-    val density = betaDensity(logistic) + logisticJacobian.log
-    RandomVariable(logistic, density)
-  }
+  override def supportTransform(v: Variable): Real =
+    Real.one / (Real.one + (v * -1).exp)
+  def supportTransformLogJacobian(v: Variable): Real =
+    supportTransform(v).log + (1 - supportTransform(v)).log
 
   val generator: Generator[Double] =
     Gamma(a, 1).generator.zip(Gamma(b, 1).generator).map {
@@ -196,7 +202,12 @@ object Uniform {
   val beta11 = Beta(1, 1)
   val standard: Continuous = new Continuous {
     def realLogDensity(real: Real): Real = beta11.realLogDensity(real)
-    def param: RandomVariable[Real] = beta11.param
+
+    def supportTransform(v: Variable): Real =
+      beta11.supportTransform(v)
+    def supportTransformLogJacobian(v: Variable): Real =
+      beta11.supportTransformLogJacobian(v)
+
     val generator: Generator[Double] =
       Generator.from { (r, n) =>
         r.standardUniform
@@ -208,12 +219,17 @@ object Uniform {
 }
 
 /**
-  * A mixture distribution composed of Betas
+  * A mixture distribution composed of continuous distributions
   *
   * @param pmf A Map with keys representing component beta distributions and values corresponding to the probabilities of those components.
-  *            The probabilities should sum to 1.0, this is not checked for.
+  * @param _supportTransform A function mapping a real-valued variable to the range of support of the combined mixture.
+  * @param _supportTransformLogJacobian The jacobian (derivative with respect to v) of the _supportTransform function
   */
-final case class BetaMixture(pmf: Map[Continuous, Real]) extends Continuous {
+final case class ContinuousMixture(
+    pmf: Map[Continuous, Real],
+    _supportTransform: Variable => Real,
+    _supportTransformLogJacobian: Variable => Real)
+    extends Continuous {
   def realLogDensity(t: Real): Real =
     Real
       .sum(pmf.toList.map {
@@ -227,20 +243,15 @@ final case class BetaMixture(pmf: Map[Continuous, Real]) extends Continuous {
       d.generator
     }
 
-  override def param: RandomVariable[Real] = {
-    val x = new Variable
+  override def supportTransform(v: Variable): Real = _supportTransform(v)
+  override def supportTransformLogJacobian(v: Variable): Real =
+    _supportTransformLogJacobian(v)
 
-    val logistic = Real.one / (Real.one + (x * -1).exp)
-    val logisticJacobian = logistic * (1 - logistic)
+}
 
-    val logDensity = logisticJacobian.log +
-      Real
-        .sum(pmf.toList.map {
-          case (dist, prob) =>
-            (dist.realLogDensity(logistic) + prob.log).exp
-        })
-        .log
-
-    RandomVariable(logistic, logDensity)
-  }
+object ContinuousMixture {
+  def apply(pmfs: Seq[(Continuous, Real)]): ContinuousMixture =
+    new ContinuousMixture(pmfs.toMap,
+                          pmfs.head._1.supportTransform,
+                          pmfs.head._1.supportTransformLogJacobian)
 }
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -144,7 +144,7 @@ object Exponential {
 /**
   * A Beta distribution with expectation `a/(a + b)` and variance `ab/((a + b)^2 (1 + a + b))`.
   */
-final case class Beta(a: Real, b: Real) {
+final case class Beta(a: Real, b: Real) extends Continuous {
   def realLogDensity(real: Real): Real =
     If(real >= 0,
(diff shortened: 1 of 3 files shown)


avi-stripe:
Awesome, thank you for diving into this.

As a high level comment, it feels like we should introduce a `Support` trait and have every `Continuous` distribution be able to return its `support: Support`. (This actually existed a long time ago in an earlier version of rainier).

I'll make some more detailed comments as well.

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -17,21 +17,29 @@ trait Injection { self =>
...
 
-    def param: RandomVariable[Real] = dist.param.map(forwards)
+    override def supportTransform(v: Variable): Real =
+      forwards(dist.supportTransform(v))
+
+    override def supportTransformLogJacobian(v: Variable): Real =
+      logForwardsJacobian(supportTransform(v)) + dist
+        .supportTransformLogJacobian(v)
+
+    override def param: RandomVariable[Real] = dist.param.map(forwards)
```
avi-stripe: This stuff gets confusing, but I believe you should just delete this and leave it to the default implementation, which will call `supportTransform` and `supportTransformLogJacobian`, otherwise those won't ever get called.

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -17,21 +17,29 @@ trait Injection { self =>
...
+           logBackwardsJacobian(real),
          Real.zero.log)
 
     val generator: Generator[Double] =
       Generator.require(self.requirements) { (r, n) =>
         n.toDouble(forwards(dist.generator.get(r, n)))
       }
 
-    def param: RandomVariable[Real] = dist.param.map(forwards)
+    override def supportTransform(v: Variable): Real =
```
avi-stripe: You don't need the `override` keyword if there's no parent implementation

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -206,3 +217,41 @@ object Uniform {
...
+    }
+
+  override def supportTransform(v: Variable): Real = _supportTransform(v)
+  override def supportTransformLogJacobian(v: Variable): Real =
+    _supportTransformLogJacobian(v)
+
+}
+
+object ContinuousMixture {
+  def apply(pmfs: Seq[(Continuous, Real)]): ContinuousMixture =
```
avi-stripe: This API is too implicit for me. I think if we were doing it without a separate `Support` object I'd suggest having the first component be more special. Something like:

```
def apply(first: Continuous, rest: Map[Continuous,Real]): ContinuousMixture = ...
def apply(first: Continuous, second: Continuous, pSecond: Real): ContinuousMixture = apply(first, Map(second -> pSecond))
```

Better though would be to have `support` on `Continuous` and either dynamically check that `support` matches across all of them, or (better/fancier) have a `union` operation on `Support`

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -206,3 +217,41 @@ object Uniform {
...
 }
+
+/**
+  * A mixture distribution composed of continuous distributions
+  *
+  * @param pmf A Map with keys representing component beta distributions and values corresponding to the probabilities of those components.
+  * @param _supportTransform A function mapping a real-valued variable to the range of support of the combined mixture.
+  * @param _supportTransformLogJacobian The jacobian (derivative with respect to v) of the _supportTransform function
+  */
+final case class ContinuousMixture(
```
avi-stripe: Should we just call this `Mixture` and get rid of the other one? All of the near-term applications I can think of for `Mixture` are of `Continuous` dists anyway. And it's confusing to have both of them if we don't need it.

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -17,21 +17,29 @@ trait Injection { self =>
     See https://en.wikipedia.org/wiki/Probability_density_function#Dependent_variables_and_change_of_variables
     This function should be log(d/dy backwards(y)), where y = forwards(x).
    */
-  def logJacobian(y: Real): Real
+  def logBackwardsJacobian(y: Real): Real
+  def logForwardsJacobian(x: Real): Real
 
   def transform(dist: Continuous): Continuous = new Continuous {
     def realLogDensity(real: Real): Real =
       If(isDefinedAt(real),
```
avi-stripe: random note I didn't have anywhere else to include: if we add a `Support` trait we should move `isDefinedAt` to there

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
mikeheaton-stripe: One can imagine a mixture of discrete distributions, right? That's why I specified `Continuous`. Agree the existing `Mixture` is confusing.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
mikeheaton-stripe: I like the `Support` object. A `union` operation should be easy enough if `isDefined` is a `Real => Bool` function, which we can `\/` over the constituent parts of the union.

mikeheaton-stripe: Actually I guess it would be a`Real => Real` using `If`

mikeheaton-stripe: Though that is a lot of `If` checks which might cause a slowdown or bad derivatives... wdyt?

avi-stripe: I think we can just enumerate the cases for a union b, rather than something generic based om isDefinedAt

mikeheaton-stripe: Oh, we'd possibly need a new transform for the new support as well as an `isDefined`. Damn.

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
mikeheaton-stripe: Done

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
avi-stripe: So mixtures of `Categorical` is just another `Categorical`. If you get into mixtures of discrete but unbounded dists, then yes, you'd need something special for that (like a mixture of `Poisson`). But that's quite a special case that I'm not too worried about supporting for the time being.

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
mikeheaton-stripe: I rewrote using `Support`, hopefully fixes this.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -206,3 +205,40 @@ object Uniform {
...
+}
+
+object ContinuousMixture {
+
+  /**
+    * Allows the creation of a mixture distribution without specifying the support, using the support of one of its components.
+    * @param supportDistribution A distribution (and its weight) in the mixture with support which is the union of all supports in the mixture.
+    * @param otherDistributions The other distributions, in a Map(distribution -> distr weight) form.
+    */
+  def apply(supportDistribution: (Continuous, Real),
```
avi-stripe: should we compute the support distribution's weight with 1.0 - the sum of the other dists weight?

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
```diff
@@ -0,0 +1,67 @@
...
+  */
+object OpenUnitSupport extends Support {
+  def transform(v: Variable): Real =
+    Real.one / (Real.one + (v * -1).exp)
+
+  def logJacobian(v: Variable): Real =
+    transform(v).log + (1 - transform(v)).log
+
+  // (x > 0) <=> ((x + |x|) != 0) && (x != 0) <=> (x + |x|) * x != 0
+  def greaterThan0(x: Real): Real = (x + x.abs) * x
```
avi-stripe: Is this better than `x > 0`, which I think does something similar?

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  // x is in (0, 1) <=> (x > 0) && (1 - x > 0)
+  def in01(x: Real): Real = greaterThan0(x) * greaterThan0(1 - x)
```
avi-stripe: feels like this could just be inlined into `isDefinedAt`

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
```diff
@@ -0,0 +1,67 @@
...
+  /*
+  Jacobian time: we need pdf(x) and we have pdf(f(x)) where f(x) = e^x.
+  This is pdf(f(x))f'(x) which is pdf(e^x)e^x.
+  If we take the logs we get logPDF(e^x) + x.
+   */
+  def logJacobian(v: Variable): Real =
+    v
+
+  // (x > 0) <=> ((x + |x|) != 0) && (x != 0) <=> (x + |x|) * x != 0
+  def greaterThan0(x: Real): Real = (x + x.abs) * x
```
avi-stripe: duplicated from OpenUnitSupport

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -17,21 +17,35 @@ trait Injection { self =>
...
-    def param: RandomVariable[Real] = dist.param.map(forwards)
+      def logJacobian(v: Variable): Real =
+        oldSupport.logJacobian(v) + logForwardsJacobian(oldSupport.transform(v))
+
+      def isDefinedAt(y: Real): Real = oldSupport.isDefinedAt(backwards(y))
+    }
+
+    new Continuous {
+      def realLogDensity(real: Real): Real =
+        If(newSupport.isDefinedAt(real),
```
mikeheaton-stripe: This `If` statement is causing the "monad laws with all reals" test to fail. TODO look into it.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
mikeheaton-stripe: I'm not sure if that'd be less or more confusing... we could always have two `apply` methods of course.

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
mikeheaton-stripe: Oh lol, I didn't realise that existed 👍

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
mikeheaton-stripe: Fixed & uses > <

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
mikeheaton-stripe: Fixed, using >

avi-stripe:
Ok, I've been puzzling through `Injection` and why we newly need `logBackwardsJacobian` which we didn't before. Here's something that was useful for me that I'm writing down a little bit as thinking out loud.

If we're properly trying to capture the old `param` pattern, including what happens with `Injection`, it needs to look something like this:

```scala
def param = {
   val x = new Variable
   val y = xyTransform(x)
   val logDensity = xyLogJacobian(x) + yLogDensity(y)
   val z = yzTransform(y)
   RandomVariable(z, logDensity)
}
```

That is, there are actually three versions of the parameter space:
* the raw unbounded parameter space that the sampler deals with. This is the space `x` is in.
* the (possibly) bounded, but centered around `0` parameter space that we evaluate parameter densities in. This is the space `y` is in. So far (pre-mixture), the `y` space has always been the "standard" version of every distribution, ie uniforms are always evaluated in (0,1), normals are always evaluated with mean 0 and stddev 1.
* the final space that the user will actually see the parameter value in (the space `z` is in). This is also the space that `realLogDensity` is in, ie when using the distribution as a likelihood.

Normally we don't need to worry about the jacobians for the yzTransform because we are always evaluating the density in `y` space. *However*, in the case of a mixture parameter, we are going to have to evaluate the density in `z` space, which means we do need to care about the `yz` jacobians...

avi-stripe:
To be clear, I think what you're doing with `Injection` and `logBackwardsJacobian` etc is probably (or at least possibly :) correct, but it seems very expensive for the non-mixture case compared to the old way where we just did `dist.param.map(forwards)`.

mikeheaton-stripe:
To check I understand, is `yztransform` the transform that `Injection`s implement (scale, translate)?
 
So I take this to mean we should keep the `Support` the same for the transformed distribution, ie `Support` objects refer to the supports for "standard" distributions in `y`-space. EG A `Uniform(1,2)` would still use the `OpenUnitSupport`.

avi-stripe:
closing this in favor of the newer work going on in other PRs

PR #166, opened 2018-06-27
Add record and replay to RandomVariable

mio-stripe:
Add `Recording` type to store sample from the posterior and `record` and `replay` to `RandomVariable`. (More directly) addresses https://github.com/stripe/rainier/issues/135

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
@@ -1,0 +1,35 @@
+package com.stripe.rainier.compute
+
+import java.io._
+
+trait Recording {
+  def save(path: String): Unit
+  def samples: List[Array[Double]]
+}
+
+case class RecordedSample(params: List[List[Double]]) extends Recording {
+
+  def save(path: String): Unit = {
+    val file = new File(path)
+
+    /** Make directory and an empty file if they don't exist **/
+    (new File(file.getCanonicalFile().getParentFile().toString)).mkdirs()
+    file.createNewFile()
+
+    val outputStream = new ObjectOutputStream(new FileOutputStream(file))
+    outputStream.writeObject(params)
+    outputStream.close()
+  }
+
+  def samples: List[Array[Double]] = params.map(_.toArray)
+}
+
+object RecordedSample {
+
+  def load(path: String): RecordedSample = {
+    val inputStream = new ObjectInputStream(new FileInputStream(path))
+    val samples = inputStream.readObject.asInstanceOf[List[List[Double]]]
+    inputStream.close
+    RecordedSample(samples)
+  }
+}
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
```diff
@@ -0,0 +1,35 @@
+package com.stripe.rainier.compute
+
+import java.io._
+
+trait Recording {
```
avi-stripe: why make this a trait instead of just having the case class be called `Recording` for now? As in, what alternative implementations of the trait do you imagine there will be?

rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
```diff
@@ -0,0 +1,35 @@
...
+  def save(path: String): Unit
```
avi-stripe: might be useful to break this into a `serialize` that produces a stream of bytes and a `save` that writes them to a file (eg, so that we can stash the bytes in thrift or something instead in a hadoop context)

rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
```diff
@@ -0,0 +1,35 @@
...
+    outputStream.writeObject(params)
+    outputStream.close()
+  }
+
+  def samples: List[Array[Double]] = params.map(_.toArray)
+}
+
+object RecordedSample {
+
+  def load(path: String): RecordedSample = {
```
avi-stripe: same comment re supporting byte arrays also

rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
mio-stripe: 👍

rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
mio-stripe: Good call. Done.

rainier-core/src/main/scala/com/stripe/rainier/compute/Recording.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  def samples: List[Array[Double]] = params.map(_.toArray)
+}
+
+object Recording {
+
+  def serialize[A](value: A): Array[Byte] = {
+    val stream = new ByteArrayOutputStream()
+    val oos = new ObjectOutputStream(stream)
+    oos.writeObject(value)
```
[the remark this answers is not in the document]
mio-stripe: Oh, nice, no I haven't. I'll give that a go.

mio-stripe:
@avi-stripe 👀 updates?

PR #168, opened 2018-06-28
Support class for Continuous Distributions

mikeheaton-stripe:
Adds the `Support` class.

As discussed, this class factors out the transforms we need for calculating `logDensity` and `param` methods on continuous distributions.

I did make `param` into a pattern because almost every instance of `Continuous` can use it, but overrode the definition for `Injection`. The injective transforms won't change the support so `injection.transform(cntsDist).support` inherits directly from `cntsDist`.

This passes the existing tests.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -23,7 +23,7 @@ trait Continuous extends Distribution[Double] {
 /**
   * A Continuous Distribution that inherits its transforms from a Support object.
   */
-trait StandardContinuous extends Continuous {
+private trait StandardContinuous extends Continuous {
   private[rainier] val support: Support
 
   def param: RandomVariable[Real] = {
@@ -44,7 +44,7 @@ trait LocationScaleFamily { self =>
   def logDensity(x: Real): Real
   def generate(r: RNG): Double
 
-  val standard: StandardContinuous = new StandardContinuous {
+  val standard: Continuous = new StandardContinuous {
     val support: Support = RealSupport
 
     val generator: Generator[Double] =
@@ -97,7 +97,7 @@ object Gamma {
   def apply(shape: Real, scale: Real): Continuous =
     standard(shape).scale(scale)
 
-  def standard(shape: Real): StandardContinuous = new StandardContinuous {
+  def standard(shape: Real): Continuous = new StandardContinuous {
     val support = PositiveSupport
 
     def realLogDensity(real: Real): Real =
@@ -143,7 +143,7 @@ object Gamma {
   * An Exponential distribution with expectation `1/rate`
   */
 object Exponential {
-  val standard: StandardContinuous = Gamma.standard(1.0)
+  val standard: Continuous = Gamma.standard(1.0)
   def apply(rate: Real): Continuous =
     standard.scale(Real.one / rate)
 }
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -8,16 +8,26 @@ import scala.annotation.tailrec
...
 
+  private[rainier] val support: Support
+
   def scale(a: Real): Continuous = Scale(a).transform(this)
   def translate(b: Real): Continuous = Translate(b).transform(this)
   def exp: Continuous = Exp.transform(this)
 
   private[rainier] def realLogDensity(real: Real): Real
+
+  def param: RandomVariable[Real] = {
```
avi-stripe: nit: rather than having this in `Continuous` and overriding it in `Injection`, can we move it down to some private `BaseContinuous` trait that everything other than `Injection` can inherit from? I prefer to avoid overrides if at all possible.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -8,16 +8,26 @@ import scala.annotation.tailrec
...
+    val x = new Variable
+
+    val paramSupport = support.transform(x)
```
avi-stripe: nit: can we call this something like `transformed` instead?

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
avi-stripe: We could also move `support` down there for now I think

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
```diff
@@ -0,0 +1,44 @@
+package com.stripe.rainier.core
+
+import com.stripe.rainier.compute._
+
+/**
+  * A trait for objects representing the support of a continuous distribution.
+  * Specifies a function to transform a real-valued variable to this range,
+  * and its log-jacobian.
+  */
+trait Support { self =>
```
avi-stripe: you can leave off the `self` here since we never use it

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -27,17 +44,15 @@ trait LocationScaleFamily { self =>
   def logDensity(x: Real): Real
   def generate(r: RNG): Double
 
-  val standard: Continuous = new Continuous {
+  val standard: StandardContinuous = new StandardContinuous {
```
avi-stripe: the type annotation here should just be `Continuous` to keep `StandardContinuous` private

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -18,6 +16,25 @@ trait Continuous extends Distribution[Double] { self =>
...
 
   private[rainier] def realLogDensity(real: Real): Real
+
+  def param: RandomVariable[Real]
+}
+
+/**
+  * A Continuous Distribution that inherits its transforms from a Support object.
+  */
+trait StandardContinuous extends Continuous {
```
avi-stripe: private

mikeheaton-stripe:
Did those and changed a couple of other similar refs.

PR #169, opened 2018-06-29
Injection maps handle Supports

mikeheaton-stripe:
This PR makes the `Injection` transforms take care of the `Support` of distributions.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -100,7 +100,7 @@ object Gamma {
     standard(shape).scale(scale)
 
   def standard(shape: Real): StandardContinuous = new StandardContinuous {
-    val support = new BoundedBelowSupport(Real.zero)
+    val support = BoundedBelowSupport(Real.zero)
 
     def realLogDensity(real: Real): Real =
       If(real > 0,
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -8,6 +8,8 @@ import scala.annotation.tailrec
   * A Continuous Distribution, with method `param` allowing conversion to a RandomVariable.
   */
 trait Continuous extends Distribution[Double] {
+  private[rainier] val support: Support
+
   def logDensity(t: Double): Real =
     realLogDensity(Real(t))
 
@@ -24,8 +26,6 @@ trait Continuous extends Distribution[Double] {
   * A Continuous Distribution that inherits its transforms from a Support object.
   */
 trait StandardContinuous extends Continuous {
-  private[rainier] val support: Support
-
   def param: RandomVariable[Real] = {
     val x = new Variable
 
@@ -45,7 +45,7 @@ trait LocationScaleFamily { self =>
   def generate(r: RNG): Double
 
   val standard: StandardContinuous = new StandardContinuous {
-    val support: Support = RealSupport
+    val support: Support = UnboundedSupport
 
     val generator: Generator[Double] =
       Generator.from { (r, n) =>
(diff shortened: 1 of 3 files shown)


avi-stripe:
:+1: to what's here so far

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -6,19 +6,21 @@ import com.stripe.rainier.unused
...
   * Injective transformations
   */
-trait Injection { self =>
+private[rainier] trait Injection { self =>
   def forwards(x: Real): Real
   def backwards(y: Real): Real
   def isDefinedAt(@unused y: Real): Real = Real.one
   def requirements: Set[Real]
-
+  def supportTransform(supp: Support): Support
```
avi-stripe: can we call this method `transformSupport` instead?

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -35,37 +37,56 @@ trait Injection { self =>
...
-  * Class to scale a distribution under multiplication by a scale factor
+  * Class to scale a distribution under multiplication by a positive scale factor.
+  * We assume that (a > 0).
   */
 final case class Scale(a: Real) extends Injection {
-  private val lj = a.log * -1
   def forwards(x: Real): Real = x * a
   def backwards(y: Real): Real = y / a
-  def logJacobian(y: Real): Real = lj
+  def logJacobian(y: Real): Real = a.log * -1
```
avi-stripe: Is there a reason you got rid of the shared object for this? I try to be careful about not over-allocating where possible.

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
```diff
@@ -35,37 +37,56 @@ trait Injection { self =>
...
-  //this is rarely important because it depends solely on y, which is usually data and not parameters
   def logJacobian(y: Real): Real = y.log * -1
 
-  override def isDefinedAt(y: Real): Real =
-    y > 0
+  override def isDefinedAt(y: Real): Real = y > 0
   val requirements: Set[Real] = Set.empty
+
+  def supportTransform(supp: Support): Support = supp match {
+    case UnboundedSupport         => UnboundedSupport
```
avi-stripe: shouldn't this end up as `BoundedBelowSupport(Real.zero)`?

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -152,7 +154,7 @@ object Exponential {
   * A Beta distribution with expectation `a/(a + b)` and variance `ab/((a + b)^2 (1 + a + b))`.
   */
 final case class Beta(a: Real, b: Real) extends StandardContinuous {
-  val support = OpenUnitSupport
+  val support = new BoundedSupport(Real.zero, Real.one)
```
avi-stripe: don't need `new` for a case class

mikeheaton-stripe: Done

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
mikeheaton-stripe: Changed

rainier-core/src/main/scala/com/stripe/rainier/core/Injection.scala
mikeheaton-stripe: `lj` wasn't being used anywhere else so I thought it should be inlined... what do you mean by over-allocating here?

mikeheaton-stripe: Ah, putting it in a `val` will make it calculate only once, right?

avi-stripe: exactly

PR #170, opened 2018-06-29
Placeholder

avi-stripe:
This is a refactoring preparing the way for mini-batch style computation. The primary change is that it changes `Distribution` from being parameterized as `Distribution[T]` to `Distribution[T,P]`, where `P` is now a "placeholder type" that is in `Real`-space instead of concrete number space. For example, what would previously have been `Distribution[Double]` is now `Distribution[Double,Real]`, and what would have previously been `Distribution[(Int,Int)]` is now `Distribution[(Int,Int),(Real,Real)]`.

The `logDensity` method is now defined in terms of the `P` type instead of the `T` type. (This was already effectively true for `Continuous` with `realLogDensity`, so in some sense this is just a generalization of that pattern).

The relationship between any given `T,P` pair is managed by a `Placeholder[T,P]` typeclass, which should generally be implicitly inferrable through a combination of primitive mappings (like `(Int,Real)` and `(Double,Real)`) and composition rules. Currently, `Placeholder` handles the `T->P` conversion (necessary for taking the `logDensity` of a `T` value), and the `P->T` conversion, which is used for producing sampler output (in this role, `Placeholder` subsumes some of what used to be provided by `Sampleable`).

In the future, `Placeholder` will also need to create and manage `Variable`-space versions of the placeholder objects, but that will introduce a lot more type complexity and will be done as a follow-on PR.

This PR ignores `Predictor` (which is not a `Distribution`), though we will eventually need to make it `Placeholder`-aware as well.

cc @mio-stripe @roban-stripe @mikeheaton-stripe @andyscott-stripe for possible reviews

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
@@ -45,8 +45,8 @@ final case class Categorical[T](pmf: Map[T, Real])
   def toMixture[V, P](implicit ev: T <:< Distribution[V, P],
                       placeholder: Placeholder[V, P]): Mixture[V, P, T] =
     Mixture[V, P, T](pmf)
-  def toMultinomial: Predictor[Int, Real, Map[T, Int], Multinomial[T]] =
-    Predictor.from[Int] { k =>
+  def toMultinomial: Predictor[Int, Map[T, Int], Multinomial[T]] =
+    Predictor.from { k: Int =>
       Multinomial(pmf, k)
     }
 }
@@ -55,8 +55,8 @@ object Categorical {
 
   def boolean(p: Real): Categorical[Boolean] =
     Categorical(Map(true -> p, false -> (Real.one - p)))
-  def binomial(p: Real): Predictor[Int, Real, Int, Binomial] =
-    Predictor.from[Int] { k =>
+  def binomial(p: Real): Predictor[Int, Int, Binomial] =
+    Predictor.from { k: Int =>
       Binomial(p, k)
     }
 
@@ -106,7 +106,7 @@ final case class Multinomial[T](pmf: Map[T, Real], k: Real)
   */
 final case class Binomial(p: Real, k: Real) extends Distribution[Int, Real] {
   val multi: Multinomial[Boolean] =
-    Categorical.boolean(p).toMultinomial(k)
+    Multinomial(Map(true -> p, false -> (1 - p)), k)
 
   def generator: Generator[Int] = {
     val poissonGenerator = Poisson(p * k).generator.map { _.min(k) }
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
@@ -7,7 +7,8 @@ import com.stripe.rainier.compute.{If, Real}
   *
   * @param pmf A map with keys corresponding to the possible outcomes and values corresponding to the probabilities of those outcomes
   */
-final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T,Map[T,Real]] {
+final case class Categorical[T](pmf: Map[T, Real])
+    extends Distribution[T, Map[T, Real]] {
   def map[U](fn: T => U): Categorical[U] =
     flatMap { t =>
       Categorical(Map(fn(t) -> Real.one))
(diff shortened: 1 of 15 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -24,8 +25,14 @@ final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T] {
         .map { case (u, ups) => u -> Real.sum(ups.map(_._2)) }
         .toMap)
 
-  def logDensity(t: T): Real =
-    pmf.getOrElse(t, Real.zero).log
+  //there should be exactly one value set to 1.0
```
mio-stripe: So rather than computing `logDensity` of a `t: T`, we compute `logDensity` of a `Map[T, Real]` of the form `Map(t -> Rea.one)`, right?

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -32,7 +27,7 @@ trait LocationScaleFamily { self =>
       Generator.from { (r, n) =>
         generate(r)
       }
-    def realLogDensity(real: Real): Real =
+    def logDensity(real: Real): Real =
```
roban-stripe: What's the advantage of this `self` pattern over just making this an abstract `trait` with an unimplemented `logDensity` method? When I read about this pattern it seemed to be used when you wanted to specify a type on the `self` parameter.

avi-stripe: This is a bit strange - the outer `LocationScaleFamily` here is a trait for, basically, factory objects that produce new instances of `Continuous`. So in this case the `self` is just so that the definition of `logDensity` inside the definition of this new `Continuous` object can refer to the outer factory object (whereas `this` would refer to the `Continuous` object).

roban-stripe: Oh, I see. Cool!

rainier-core/src/main/scala/com/stripe/rainier/core/Distribution.scala
```diff
@@ -5,46 +5,17 @@ import com.stripe.rainier.compute.Real
 /**
   * Basic probability distribution trait
   */
-trait Distribution[T] extends Likelihood[T] { self =>
-  def logDensity(t: T): Real
-  def logDensities(list: Seq[T]): Real = Real.sum(list.map(logDensity))
+abstract class Distribution[T, P](implicit placeholder: Placeholder[T, P])
+    extends Likelihood[T] {
+  def logDensity(t: P): Real
+  def logValueDensity(t: T): Real = logDensity(placeholder.wrap(t))
```
roban-stripe: This name makes it sound to me like we're calculating the density of a log value (or at least its ambiguous). What if we called it `valueLogDensity`?

rainier-core/src/main/scala/com/stripe/rainier/core/Distribution.scala
```diff
@@ -5,46 +5,17 @@ import com.stripe.rainier.compute.Real
...
-    // This is Gergő Nemes' approximation to the log Gamma function, plus a trick taken from Boost's lgamma function:
-    // since the Nemes approximation isn't very accurate for small z, we instead calculate LogGamma(z + 1) - Log(z).
-    // See https://en.wikipedia.org/wiki/Stirling%27s_approximation and
-    // https://www.boost.org/doc/libs/1_50_0/libs/math/doc/sf_and_dist/html/math_toolkit/special/sf_gamma/lgamma.html.
-    val v = z + 1
-    val w = v + (Real.one / ((12 * v) - (Real.one / (10 * v))))
-    (Real(Math.PI * 2).log / 2) - (v.log / 2) + (v * (w.log - 1)) - z.log
-  }
+  def fit(t: T): RandomVariable[Distribution[T, P]] =
+    RandomVariable(this, logDensity(placeholder.wrap(t)))
```
roban-stripe: Why not use `logValueDensity` (or `valueLogDensity`) here?

avi-stripe: No reason, I wrote this before I wrote that method :) Will fix.

rainier-core/src/main/scala/com/stripe/rainier/core/Distribution.scala
avi-stripe: Makes sense, will fix.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
mio-stripe: Gotcha.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -112,10 +120,11 @@ final case class Binomial(p: Real, k: Int) extends Distribution[Int] {
...
   */
-final case class Mixture[T, D](pmf: Map[D, Real])(
-    implicit ev: D <:< Distribution[T])
-    extends Distribution[T] {
-  def logDensity(t: T): Real =
+final case class Mixture[T, P, D](pmf: Map[D, Real])(
+    implicit ev: D <:< Distribution[T, P],
+    placeholder: Placeholder[T, P])
+    extends Distribution[T, P] {
+  def logDensity(t: P): Real =
```
mio-stripe: nit: `p: P`

rainier-core/src/main/scala/com/stripe/rainier/core/Distribution.scala
mio-stripe: I see some `Real.sum(...)`s in tests, do we want a `valueLogDensities`, too?

mio-stripe:
Cool, left a couple tiny comments, but LGTM

avi-stripe:
@mio-stripe @roban-stripe I've rebased and addressed comments. Will merge if CI passes.

PR #171, opened 2018-06-30
Add Mixture Distribution

mikeheaton-stripe:
Adds the ContinuousMixture distribution itself, along with a supporting `union` method to `Support`.

WIP: currently passes one test but fails another. Debugging and extra tests to come.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
@@ -41,9 +41,6 @@ final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T] {
     }
   }
 
-  def toMixture[V](implicit ev: T <:< Distribution[V]): Mixture[V, T] =
-    Mixture[V, T](pmf)
-
   def toMultinomial: Predictor[Int, Map[T, Int], Multinomial[T]] =
     Predictor.from { k: Int =>
       Multinomial(pmf, k)
@@ -129,28 +126,6 @@ final case class Binomial(p: Real, k: Int) extends Distribution[Int] {
 
   def logDensity(t: Int): Real =
     multi.logDensity(Map(true -> t, false -> (k - t)))
-}
-
-/**
-  * A Mixture distribution
-  *
-  * @param pmf A Map with keys representing component distributions and values corresponding to the probabilities of those components
-  */
-final case class Mixture[T, D](pmf: Map[D, Real])(
-    implicit ev: D <:< Distribution[T])
-    extends Distribution[T] {
-  def logDensity(t: T): Real =
-    Real
-      .sum(pmf.toList.map {
-        case (dist, prob) =>
-          (ev(dist).logDensity(t) + prob.log).exp
-      })
-      .log
-
-  def generator: Generator[T] =
-    Categorical(pmf).generator.flatMap { d =>
-      d.generator
-    }
 }
 
 /**
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
@@ -41,6 +41,9 @@ final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T] {
     }
   }
 
+  def toMixture[V](implicit ev: T <:< Continuous): Mixture =
+    Mixture(pmf.map{ case (k, v) => (ev(k), v) })
+
   def toMultinomial: Predictor[Int, Map[T, Int], Multinomial[T]] =
     Predictor.from { k: Int =>
       Multinomial(pmf, k)
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
@@ -210,7 +210,7 @@ object Uniform {
     standard.scale(to - from).translate(from)
 }
 
-class ContinuousMixture(components: Map[Continuous, Real])
+case class ContinuousMixture(components: Map[Continuous, Real])
     extends StandardContinuous {
   def generator: Generator[Double] =
     Categorical(components).generator.flatMap { d =>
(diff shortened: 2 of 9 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
```diff
@@ -8,9 +8,41 @@ import com.stripe.rainier.compute._
...
+      BoundedAboveSupport(realMax(thisMax, thatMax))
+    case (BoundedSupport(thisMin, thisMax), BoundedSupport(thatMin, thatMax)) =>
+      BoundedSupport(realMin(thisMin, thatMin), realMax(thisMax, thatMax))
+  }
+
+  def isDefinedAt(real: Real): Real
+}
+
+object Support {
+  private def realMin(a: Real, b: Real): Real = ((a - b).abs - (b + a)) / 2.0
```
avi-stripe: Neat. Should we move this into `Real` itself?

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -211,3 +209,25 @@ object Uniform {
...
+    }
+
+  val support = Support.union(components.keys.map { _.support })
+
+  def realLogDensity(real: Real): Real =
+    Real
+      .sum(
+        components.map {
+          case (dist, weight) =>
+            If(dist.support.isDefinedAt(real),
```
avi-stripe: do we need this? can't we rely on `dist.realLogDensity` doing basically the same thing?

rainier-core/src/main/scala/com/stripe/rainier/core/Support.scala
mikeheaton-stripe: Happy to!

mikeheaton-stripe:
I think this is completed. It passes all the `sbcCheck` tests, but I'm leaving them commented so that they don't run and take five minutes every time we test.

The generator tests and param tests (mostly cribbed from `ContinuousTest`) don't pass consistently. It might be sensible to remove them, I left them here in case we wanted to work on them more.

rainier-tests/src/test/scala/com/stripe/rainier/core/ContinuousMixtureTest.scala
```diff
@@ -0,0 +1,241 @@
...
+      Map(
+        Normal(-3, 1) -> 0.5,
+        Exponential(1.0) -> 0.5
+      )
+    ),
+    -1.0)
+
+  /********************************
+ * Param Tests
+ * These don't pass consistently, although learning seems to be occurring in the right direction.
```
avi-stripe: should we comment them out too, then?

rainier-tests/src/test/scala/com/stripe/rainier/core/ContinuousMixtureTest.scala
```diff
@@ -0,0 +1,241 @@
...
+import com.stripe.rainier.sampler._
+import org.scalatest.FunSuite
+
+class ContinuousMixtureTest extends FunSuite {
+  implicit val rng: RNG = ScalaRNG(1528673302081L)
+
+  def sbcCheck(description: String)(prior: Continuous): Unit = {
+    val emptyEvaluator = new Evaluator(Map.empty)
+
+    def binomialQuantile(q: Double, n: Int, p: Double): Int = {
```
avi-stripe: oh good, let's move this lower/upper quantile check inside SBC?

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -211,3 +209,36 @@ object Uniform {
   def apply(from: Real, to: Real): Continuous =
     standard.scale(to - from).translate(from)
 }
+
+case class ContinuousMixture(components: Map[Continuous, Real])
```
avi-stripe: I still think we should just rename this to `Mixture` and delete the other one.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
```diff
@@ -66,7 +64,7 @@ trait LocationScaleFamily { self =>
   */
 object Normal extends LocationScaleFamily {
   def logDensity(x: Real): Real =
-    (x * x) / -2.0
+    ((x * x) / -2.0) - 0.5 * Real(2 * math.Pi).log
```
avi-stripe: ooh I can see how this might matter in a mixture

rainier-tests/src/test/scala/com/stripe/rainier/core/ContinuousMixtureTest.scala
mikeheaton-stripe: I culted it from there! It's `private` inside SBC though, I'll unprivate it.

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
mikeheaton-stripe: Done

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
mikeheaton-stripe: When I spotted that I thought for about half a minute that the whole of Rainier was out by a constant factor...

rainier-tests/src/test/scala/com/stripe/rainier/core/ContinuousMixtureTest.scala
mikeheaton-stripe: They are commented out!

rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala
mikeheaton-stripe: Actually before I do... I'm just worried about removing the ability to have `Mixture`s for non-`Continuous` distributions. At the moment, `Categorical.toMixture` can work with any type of distribution.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -41,9 +41,6 @@ final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T] {
     }
   }
 
-  def toMixture[V](implicit ev: T <:< Distribution[V]): Mixture[V, T] =
```
avi-stripe: You can still have this but where `ev: T <:< Continuous`

mikeheaton-stripe: Kk, done

avi-stripe:
ship it!

PR #173, opened 2018-07-05
Placeholder (take 2)

avi-stripe:
This is a revision of https://github.com/stripe/rainier/pull/170 that keeps the parameterization of `Distribution[T]` the same by playing a bunch of games with typeclasses and implicits in `Likelihood.scala`


avi-stripe:
cc @roban-stripe

PR #175, opened 2018-07-10
Rectifier

avi-stripe:
* Adds RectifierOp
* Uses it in the implementations of comparison ops
* Uses comparison ops for min/max
* Uses max to do the logsumexp trick


avi-stripe:
We also lurked in some jacobian fixes.

mio-stripe:
Nice! LGTM modulo a CI failure.

avi-stripe:
looks from the tests like we broke Exponential somehow?

mikeheaton-stripe:
I'll look into jt

PR #177, opened 2018-07-20
rerun benchmark script on master

mio-stripe:
I just re-ran the benchmark script on the latest version of master to have an current version to compare to.


avi-stripe:
LGTM

PR #178, opened 2018-07-20
Add l1 and l2 error benchmarks

mio-stripe:
This adds a couple new benchmarks that will be useful for evaluating the Map => ListMap change.


avi-stripe:
LGTM

avi-stripe:
Actually,  before we merge this, can we add both 100 and 1000 data point variations? I'd like to have both baselines.

PR #179, opened 2018-07-23
Disregard version format for shaded asm jar

andyscott:
This should allow us to publish the infrequently updated shaded asm jar.


mikeheaton-stripe:
lgtm

Issue #116, opened 2018-06-04
Switch tests to use SBC


avi-stripe:
Once https://github.com/stripe/rainier/pull/107 is merged, we should use SBC for our regression tests. It's impractical to run them each time so we'll need some scheme that involves manually running them any time the sampler behavior changes and then storing a "hash" (eg, the first K samples, or the rHat of the first run) to verify that it hasn't.

PR #182, opened 2018-07-31
Use ListMap instead of Map for deterministic ordering of variables.

mio-stripe:
This is a change to switch to testing a la SBC https://github.com/stripe/rainier/issues/116


avi-stripe:
Code changes look fine to me; as discussed, we shouldn't merge yet.
We should also include a benchmarks diff.

mio-stripe:
I got the prs stacked properly now: we'll merge the SBC prs down into this branch and merge this last.

rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
```diff
@@ -104,12 +104,12 @@ Because it is common for ax to have a large number of terms, this is deliberatel
 as equality comparisons would be too expensive. The impact of this is subtle, see [0] at the bottom of this file
 for an example.
  */
-private final class Line private (val ax: Map[NonConstant, BigDecimal],
+private final class Line private (val ax: ListMap[NonConstant, BigDecimal],
```
avibryant, line 108: Along with the change to ListMap, I think we should change Line to a case class (so that it uses structural equality instead of identity equality). Incidentally I checked the equality semantics on ListMap and they're the same as Map, ie order isn't considered - which I think is good, for our purposes.

avibryant: We can leave this for the capped size listmap change.

PR #183, opened 2018-07-31
Narrow the use of Context

avi-stripe:

This makes some very minor changes to limit the places where we use `Context` so that we can change its API later.


mio-stripe:
LGTM

PR #185, opened 2018-07-31
compileDensity

avi-stripe:

Add a simple `compileDensity` convenience method to `Context` and use it from `Walkers`


mio-stripe:
LGTM

PR #186, opened 2018-07-31
PartialEvaluator

avi-stripe:

Introduce a `PartialEvaluator` which is like `Evaluator` but which takes maps `Variable` to `Real` instead of `Double`; this lets us simplify a compute graph when knowing the value of some but not all variables.


mio-stripe:
LGTM

PR #188, opened 2018-07-31
rename BoxedReal to Target

avi-stripe:

Renames `BoxedReal` to `Target` and splits it out to its own file in anticipation of it becoming a richer object later on.


mio-stripe:
LGTM

PR #189, opened 2018-07-31
add batches param to sample()

avi-stripe:

Adds a `batches` param to `sample()` and related methods, with a default value of 1, that currently does not do anything.
Adds a `BatchNormal` example that sets it.


mio-stripe:
LGTM

PR #190, opened 2018-07-31
Shuffle around `Fittable` and `Likelihood`

avi-stripe:

Depends on https://github.com/stripe/rainier/pull/188 .

This does some renaming and minor associated refactoring. In particular:

- the base trait for objects that you can call `fit` on is now `Fittable` (this used to be `Likelihood`)
- the wrapper class used to implicitly provide the `fit` methods is now `Fittable.Ops` (it used to be `Likelihood.Fittable`
- the typeclass trait for defining the likelihood function for a fittable object is now `Likelihood` (it used to be `Likelihood.Fn`
- `Likelihood` now uses `target(L,T): Target` instead of `apply(L,T): Real` as its main likelihood method
- `RandomVariable.fit(l, t)` is now a more explicit alternative to calling `l.fit(t)` that doesn't require the `Ops` implicit


mio-stripe:
LGTM

PR #192, opened 2018-08-02
[wip] discrete distributions/mixtures

DavidRdgz:
[wip] feedback regarding the discrete distributions would be greatly appreciated. For example, the negative binomial, samples grow linearly with n. Should the ZeroInflatedPoisson and ZeroInflatedNegativeBinomial distributions be removed? Is the use of `support`, here, possibly confusing (since the continuous dists use this to mean something else)? Lastly, is there an preferred way to create unit tests for a mixture of discrete distributions?

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
@@ -6,10 +6,10 @@ trait Discrete extends Distribution[Int] { self: Discrete =>
   def logDensity(v: Real): Real
 
   def zeroInflated(psi: Real) =
-    DiscreteMixture(Map(DiscreteConstant(0.0) -> (1 - psi), self -> psi))
+    constantInflated(0.0, psi)
 
   def constantInflated(constant: Real, psi: Real) =
-    DiscreteMixture(Map(DiscreteConstant(constant) -> (1 - psi), self -> psi))
+    DiscreteMixture(Map(DiscreteConstant(constant) -> psi, self -> (1 - psi)))
 }
 
 object Discrete {
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
@@ -2,8 +2,14 @@ package com.stripe.rainier.core
 
 import com.stripe.rainier.compute.{If, Real}
 
-trait Discrete extends Distribution[Int] {
+trait Discrete extends Distribution[Int] { self: Discrete =>
   def logDensity(v: Real): Real
+
+  def zeroInflated(psi: Real) =
+    DiscreteMixture(Map(DiscreteConstant(0.0) -> (1 - psi), self -> psi))
+
+  def constantInflated(constant: Real, psi: Real) =
+    DiscreteMixture(Map(DiscreteConstant(constant) -> (1 - psi), self -> psi))
 }
 
 object Discrete {
@@ -24,9 +30,8 @@ final case class DiscreteConstant(constant: Real) extends Discrete {
       n.toInt(constant)
     }
 
-  def logDensity(v: Real): Real = {
+  def logDensity(v: Real): Real =
     If(v - constant, Real.negInfinity, 0)
-  }
 }
 
 /**
@@ -64,25 +69,40 @@ final case class Geometric(p: Real) extends Discrete {
 }
 
 /**
-  * Negative Binomial distribution with expectation `n(1-p)/p`
+  * Negative Binomial distribution with expectation `n*p/(1-p)`
   *
   * @param n Total number of failures
   * @param p Probability of success
   */
 final case class NegativeBinomial(n: Real, p: Real) extends Discrete {
-  val generator: Generator[Int] =
-    Generator.require(Set(n, p)) { (r, m) =>
+  val generator: Generator[Int] = {
+    val nbGenerator = Generator.require(Set(n, p)) { (r, m) =>
       (1 to m.toInt(n))
         .map({ x =>
-          Geometric(p).generator.get(r, m)
+          Geometric(1 - p).generator.get(r, m)
         })
         .sum
     }
+    val normalGenerator =
+      Normal(n * p / (1 - p), (n * p).pow(1.0 / 2.0) / (1 - p)).generator
+        .map(_.toInt.max(0))
+
+    Generator.from {
+      case (r, m) =>
+        val pDouble = m.toDouble(p)
+        val nDouble = m.toDouble(n)
+        if (pDouble < -100 / nDouble + 1 && pDouble > 100 / nDouble - .25) {
+          normalGenerator.get(r, m)
+        } else {
+          nbGenerator.get(r, m)
+        }
+    }
+  }
 
   def logDensity(v: Real) =
     Combinatorics.factorial(n + v - 1) - Combinatorics.factorial(v) -
       Combinatorics.factorial(n - 1) +
-      n * p.log + v * (1 - p).log
+      n * (1 - p).log + v * p.log
 }
 
 /**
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -13,6 +13,78 @@ object Discrete {
...
+  *
+  * @param constant The integer value of the point mass
+  */
+final case class DiscreteConstant(constant: Real) extends Discrete {
+  val generator: Generator[Int] =
+    Generator.require(Set(constant)) { (c, n) =>
+      n.toInt(constant)
+    }
+
+  def logDensity(v: Real): Real = {
```
avi-stripe: style nit: we don't need braces around this method body

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -95,3 +167,54 @@ final case class BetaBinomial(a: Real, b: Real, k: Real) extends Discrete {
...
+      })
+}
+
+/**
+  * Zero Inflated Poisson with expecation `psi*lambda`
+  *
+  * @param psi    The probability of non-zero count
+  * @param lambda The expected (possibly non-zero) count value
+  */
+final case class ZeroInflatedPoisson(psi: Real, lambda: Real) extends Discrete {
```
avi-stripe: Rather than having special classes for these, why not just have a `zeroInflated(psi: Real)` combinator on `Discrete`?

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -13,6 +13,78 @@ object Discrete {
...
+
+/**
+  * Negative Binomial distribution with expectation `n(1-p)/p`
+  *
+  * @param n Total number of failures
+  * @param p Probability of success
+  */
+final case class NegativeBinomial(n: Real, p: Real) extends Discrete {
+  val generator: Generator[Int] =
+    Generator.require(Set(n, p)) { (r, m) =>
```
avi-stripe: As you point out, this is linear in n. For large enough n there's a normal approximation; could we switch to that after some threshold?

avi-stripe:
Ah, I didn't answer your question about unit tests. In general, we're hoping to move testing of distribution implementations over to `SBC`. I think it's fine to wait until we have that in place before adding tests for the mixture.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
DavidRdgz: Perfect. How about I add `constantInflated(constant: Real, psi: Real)`? Also, would you be ok if I change the threshold in the unit tests from within 5% to 10% for the expected values? Below, I changed the sampling from `1000` to `10000` to get the convergence. It might be easier to keep `1000` to keep the tests from taking too long.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
DavidRdgz: definitely

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
DavidRdgz: Here's my proposed thresholds for when to switch to a normal distribution: https://gist.github.com/DavidRdgz/bc313a8ef3133a082c7ca10503824e2d

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
avi-stripe: Yep, sounds good on all counts. (Though I wonder if `zeroInflated` is enough of a special case to deserve its own convenience method.)

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
avi-stripe: :+1:

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -1,9 +1,15 @@
 package com.stripe.rainier.core
 
-import com.stripe.rainier.compute.Real
+import com.stripe.rainier.compute.{If, Real}
 
-trait Discrete extends Distribution[Int] {
+trait Discrete extends Distribution[Int] { self: Discrete =>
```
DavidRdgz: Is there any objections to this `self: Discrete =>` reference?

avi-stripe: Nope, totally fine. (Would also be fine without the type annotation).

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -2,8 +2,14 @@ package com.stripe.rainier.core
 
 import com.stripe.rainier.compute.{If, Real}
 
-trait Discrete extends Distribution[Int] {
+trait Discrete extends Distribution[Int] { self: Discrete =>
   def logDensity(v: Real): Real
+
+  def zeroInflated(psi: Real) =
```
avi-stripe: this could just call `constantInflated`

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -2,8 +2,14 @@ package com.stripe.rainier.core
...
+    DiscreteMixture(Map(DiscreteConstant(0.0) -> (1 - psi), self -> psi))
+
+  def constantInflated(constant: Real, psi: Real) =
+    DiscreteMixture(Map(DiscreteConstant(constant) -> (1 - psi), self -> psi))
```
avi-stripe: Is this backwards? I would expect `psi` to be the probability of the constant.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -64,25 +69,40 @@ final case class Geometric(p: Real) extends Discrete {
...
-          Geometric(p).generator.get(r, m)
+          Geometric(1 - p).generator.get(r, m)
         })
         .sum
     }
+    val normalGenerator =
+      Normal(n * p / (1 - p), (n * p).pow(1.0 / 2.0) / (1 - p)).generator
+        .map(_.toInt.max(0))
+
+    Generator.from {
```
avi-stripe: would be slightly better as `normalGenerator.zip(nbGenerator).map{...}`

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
DavidRdgz: Ah, for consistency with Lambert ,et. al., you're right. I was thinking, "the probability of non-zero" when I wrote it. And that the expression for the expected value was easier `psi*meanOfTheNonZeroDist`. It will turn into `(1-psi)*meanOfTheNonZeroDist`.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
DavidRdgz: yes, definitely better

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
DavidRdgz:
```
    normalGenerator.zip(nbGenerator).map {
      case (normal, nbinom) =>
        val pDouble = ??
        val nDouble = ??

        if (pDouble < -100 / nDouble + 1 && pDouble > 100 / nDouble - .25) {
          normal
        } else { nbinom }
    }
```
Something like this? I'm not sure how to convert the `p` and `n` to double.

avi-stripe: ahh good point. Ok, let's leave it as is for now then.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
avi-stripe: Even apart from consistency with anything external, I think it's just much more intuitive if when you're calling `constantInflated(k, psi)`, the `k` and `psi` are directly related. So I'd rather we switched it to that.

avi-stripe:
LGTM.

PR #195, opened 2018-08-15
Use placeholders from Predictor

avi-stripe:
This changes `Predictor` to expect that its covariate -> noise-distribution function takes inputs in `Real`-space (by way of a `Placeholder`) instead of the raw representation of the covariate.


mio-stripe:
LGTM
Not sure what the build failure is about. I guessed that something wasn't updated in a test, but the tests seem to run fine.

PR #196, opened 2018-08-17
Standardize shaded-asm version

travisbrown:
I just ran into a somewhat confusing issue where I'd run `sbt +publishLocal` on rainier, and a bazel build then picked up the local `rainier-shaded-asm` artifact, which had a jar with a different hash, etc.

This change moves the `6.0` suffix into the artifact ID and uses the same version as the other rainier artifacts for `rainier-shaded-asm`. I've also updated the `version.sbt` version.

I think in general it's safe to assume (and people do tend to assume) that running `sbt + publishLocal` on an sbt project won't publish non-snapshot versions into your local Ivy cache, and after this change that assumption holds for rainier.

r? @andyscott-stripe

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -204,12 +204,12 @@ lazy val rainierTrace = project.
 /* publishable project with the shaded deps */
 lazy val shadedAsm = project.
   in(file(".rainier-shaded-asm")).
-  settings(name := "rainier-shaded-asm").
+  // note: bump version suffix when the shaded asm jar needs to be
+  // republished for a particular version of the underlying asm lib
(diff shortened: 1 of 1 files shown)


andyscott-stripe (approved):
LGTM

avibryant:
This PR's been around for quite a while but isn't passing CI. @andyscott-stripe no rush but can you take a look and either fix it or close it?

avibryant:
closing in favor of #250

kailuowang:
sorry I miscommunicated. I didn't mean that #250 will fix the whole issue, just the build error in this PR.

PR #197, opened 2018-08-30
Change `Fittable[T]` back to `Likelihood[T]`

avi-stripe:
```
What did you do last week, Avi?
I added a typeclass.
What about this week?
I removed the typeclass.
```

This may seem a little circular, but thank you for bearing with me.

This reverts back to the comfortable OO hierarchy of `Likelihood[T]` as the parent of `Distribution[T]` and `Predictor[X,Y]` (which has also conveniently lost its extraneous type params), and moves the critical methods from the `Likelihood` typeclass (like `logDensity`) back into being directly implemented in that hierarchy, while still keeping the "placeholder" model of always computing densities on `Real`-space representations of values. This is mostly achieved by starting to use package-private type members to introduce associated types but keep the external type signatures clean.

At the same time, we add `fromInt` and `fromDouble` convenience methods to the `Predictor` object to avoid the messy/verbose type inference issues of the generic `from` for the common cases.


rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala
```diff
@@ -2,19 +2,26 @@ package com.stripe.rainier.core
...
 }
 
 object Likelihood {
-  def from[L, T, U](f: (L, U) => Real)(
-      implicit ph: Placeholder[T, U]): Likelihood[L, T] =
-    new Likelihood[L, T] {
-      def target(likelihood: L, value: T) =
-        new Target(f(likelihood, ph.wrap(value)))
-    }
+  case class Ops[T, L](lh: L)(implicit ev: L <:< Likelihood[T]) {
```
mio-stripe: What is the advantage to having the both `Ops` case class and the `ops` implicit def? `Ops[T]` could instead be an `implicit class` (not a case class) and we wouldn't also need `ops`, but I suspect there's a subtle difference.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -23,20 +36,20 @@ abstract class Predictor[X, Y, Z, A](implicit ev: Z <:< Distribution[Y],
...
 
-  implicit def likelihood[X, Y, Z, A](implicit lh: Likelihood[Z, Y]) =
-    new Likelihood[Predictor[X, Y, Z, A], (X, Y)] {
-      def target(predictor: Predictor[X, Y, Z, A], value: (X, Y)) =
-        lh.target(predictor(value._1), value._2)
+      val placeholder = ph
+      def create(q: Q) = fn(q)
     }
+
+  def fromInt[Y, B](fn: Real => Distribution.Aux[Y, B]): Predictor[Int, Y] =
```
mio-stripe, line 49: Nice

mio-stripe (approved):
Nice. Left a q but nothing blocking. The type member stuff is slick.

rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala
avi-stripe: I'll try it like that, it may just work. I've tried too many variations of this code to remember :)

avi-stripe: @mio-stripe yeah that worked fine, changed (and added a comment)

PR #198, opened 2018-09-01
Mapping[T,U]

avi-stripe:
This is another crack at unifying `Placeholder` and `Sampleable`.

We replace them with three new types:

`Wrapping[T,U]` knows how to lift a concrete `T` into a `Real`-space `U`. This is used by `Likelihood[T]`.

`Mapping[T,U]` extends `Wrapping[T,U]` to also know how to (given an evaluation context) reify a `Real`-space `U` into a concrete `T`.

`ToGenerator[T,U]` is a typeclass that can convert any `T` into a `Generator[U]`. There are instances defined for `ToGenerator[Generator[T],T]` (which is just the identity function), and also for `ToGenerator[U,T]` given a `Mapping[T,U]`, which makes use of the mapping's reification.

Conceptually, when we sample from a `RandomVariable[T]`, we first need to convert it to a `RandomVariable[Generator[U]]` using some `ToGenerator[T,U]`. This is done implicitly in `sample`, or can be done explicitly using `rv.toGenerator`.

Questions and TODOs:
- we need `Mapping` instances for `Seq[T]` and for tuples of size > 2. 
- should we have a `ToGenerator[Distribution[T],T]` that calls `dist.generator`? I can't decide whether this is an obvious convenience or unnecessary magic. Maybe the best thing is for it to exist but need to be imported to be used.
- Are there better, more descriptive names for `Wrapping` and `Mapping`?


mio-stripe:
Just glancing at this, it looks awesome so far. I'll try to address your questions when I review more thoroughly tomorrow.

rainier-core/src/main/scala/com/stripe/rainier/core/Mapping.scala
```diff
@@ -0,0 +1,70 @@
...
+        u.values.flatMap { x =>
+          tu.requirements(x)
+        }.toSet
+    }
+
+  def item[T]: Mapping[T, Map[T, Real]] =
+    new Mapping[T, Map[T, Real]] {
+      def wrap(t: T): Map[T, Real] = Map(t -> Real.one)
+      def get(u: Map[T, Real])(implicit n: Numeric[Real]): T =
+        u.find { case (_, r) => n.toDouble(r) > 0 }.get._1
```
mio-stripe, line 66: Is there an assumption that `u` only has one element or only one (k,v) with n.toDouble(v) > 0?

mio-stripe (approved):
Okay, I'm onboard now. To your bullet qs, I think having `ToGenerator[Distribution[T],T]` but having to import it sounds good. I cannot think of better names. It would be good to have a code comment in `Mapping.scala` that explains this a little. I could decipher the types, but would have struggled a lot without your description of what this PR is doing.

mio-stripe:
Are these actual failures or weird travis stuff again?

PR #199, opened 2018-09-04
Mapping -> Placeholder -> Target

avi-stripe:
Depends on https://github.com/stripe/rainier/pull/198 . This still has a couple of `???` to flesh out but can mostly be reviewed as is.

The big change here is that `Target`, which is how `RandomVariable` stores each term of the density function. Now instead of just wrapping a `Real`, it also keeps a (possibly empty) mapping from `Variable` nodes that appear in the `Real`, to columns of `Array[Double]` values for that variable. Each column should be the same length (call that `n`). Conceptually, if you made `n` copies of the `real` wrapped by the `Target`, and for each copy at index `i` replace, for each column, the `Variable` key in that copy with the double at index `i` if the values, then summed all of the copies, you'd have the full density term (which may contain additional, un-replaced, `Variables`, for model parameters).

`Target` implements `inlined` which performs exactly this replacement, using `PartialEvaluator`, so that we can use it just as we used to, and that's all this PR does for now.

The other change here is to extend `Wrapping` to create a `Placeholder` which provides the missing pieces to allow one of these new `Target`s to be constructed:
- it can construct a `Real`-space object that generalizes a sequence of concrete values (instead of just wrapping a single concrete value)
- it tracks the `Variable`s in that object
- given a concrete value, it can extract the `Double`s that map to each of those `Variable`s

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
@@ -1,0 +1,26 @@
+package com.stripe.rainier.compute
+
+class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
+  def inlined = inl(placeholders)
+
+  def inl(data: Map[Variable, Array[Double]]): Real =
+    if (data.isEmpty)
+      real
+    else {
+      val inlined = 0.until(nRows(data)).map { i =>
+        val row: Map[Variable, Real] = data.map {
+          case (v, a) => v -> Real(a(i))
+        }
+        PartialEvaluator.inline(real, row)
+      }
+      Real.sum(inlined.toList)
+    }
+
+  private def nRows(data: Map[Variable, Array[Double]]): Int =
+    data.head._2.size
+}
+
+object Target {
+  def apply(real: Real): Target = new Target(real, Map.empty)
+  val empty: Target = apply(Real.zero)
+}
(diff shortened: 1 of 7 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -0,0 +1,26 @@
+package com.stripe.rainier.compute
+
+class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
+  def inlined = inl(placeholders)
+
+  def inl(data: Map[Variable, Array[Double]]): Real =
+    if (data.isEmpty)
+      real
+    else {
+      val inlined = 0.until(nRows(data)).map { i =>
```
mio-stripe, line 10: A code comment similar to your pr description would help others and future-me follow this.

mio-stripe (approved):
Okay, looks good! It takes me a little while to get my head around all of types so I think some liberal code comments and/or a doc once things stabilize explaining all of them and their relationships would be helpful.

avi-stripe:
@mio-stripe this is finished now in case you want to re-review (will timeout and merge anyway by EOD tomorrow)

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
avi-stripe: Yeah I plan to add docs at the end of this whole process.

mio-stripe:
@avi-stripe 👍 merge away.

PR #200, opened 2018-09-06
SBC testing

mio-stripe:
This is a first step toward using SBC for testing. It includes the `Map => ListMap` change to make ordering of variables deterministic under the hood. The idea is that SBC is slow to run, so we'll run it once manually for each SBCModel and, if everything looks good, copy the goldset into the SBCModel itself. Then, automated tests, will just sample from the model and check that the new samples agree with the goldset. If not, the user can debug until they are happy, rerun SBC and update the goldset when the results look reasonable.


mio-stripe:
Sanity check: I was expecting these to return the same samples and they do not
```
scala> import com.stripe.rainier.core._
import com.stripe.rainier.core._

scala> import com.stripe.rainier.sampler._

scala> import com.stripe.rainier.compute._
import com.stripe.rainier.compute._

scala> implicit val rng: RNG = ScalaRNG(1528673302081L)
rng: com.stripe.rainier.sampler.RNG = ScalaRNG(1528673302081)

scala> val bin = Binomial(0.3, 20)
bin: com.stripe.rainier.core.Binomial = Binomial(Constant(0.3),Constant(20.0))

scala> RandomVariable(bin.generator).sample(10)
res0: List[Int] = List(6, 3, 10, 6, 7, 8, 4, 7, 8, 10)

scala> RandomVariable(bin.generator).sample(10)
res1: List[Int] = List(5, 5, 10, 6, 8, 6, 6, 8, 5, 9)

```

mio-stripe:
Oh, I'm clearly wrong: I have to start from the same place each time; here, the rng is just continuing from where it left off.

mio-stripe:
Okay, great, when I reset the rng in between, it does produce the same samples.

mio-stripe:
Fixed. Now two of the SBC tests pass. The other two models are very slow to run SBC on so contain made-up goldsets for the moment.

mio-stripe:
```
sbt:rainier-core> run

....

Multiple main classes detected, select one to run:

 [1] com.stripe.rainier.core.SBCExponential
 [2] com.stripe.rainier.core.SBCLaplace
 [3] com.stripe.rainier.core.SBCLogNormal
 [4] com.stripe.rainier.core.SBCUniformNormal
[info] Packaging /Users/mio/stripe/rainier/rainier-core/target/scala-2.12/rainier-core_2.12-0.1.0.jar ...

Enter number: [info] Done packaging.
3

[info] Running com.stripe.rainier.core.SBCLogNormal
Running SBC for com.stripe.rainier.core.SBCLogNormal$@70dd5180

Running simulation-based calibration.
Repetition 320/320. Estimated time remaining: 0:00:00
Samples/sec: 30928. Max rHat: 1.005. Max thinning factor: 6
99% of bins should end up between the [ and ] quantile markers

####################      [                              ]
###############           [                              ]
########################  [                              ]
##############            [                              ]
#####################     [                              ]
################          [                              ]
###################       [                              ]
#####################     [                              ]

new goldset:
List(1.9444409266417129, 4.884477223783541, 3.166384786757497, 1.340419862750101, 62.45333008770111, 21.260796478808622, 2.145416458600115, 0.5978183569453146, 2.745814563529916, 2.326159027201714)
If this run looks good, please update the goldset in your SBCModel
```

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
```diff
@@ -15,10 +15,15 @@ final case class SBC[L, T](priorGenerators: Seq[Generator[Double]],
...
+  def posteriorSamples(nSamples: Int)(implicit rng: RNG) =
+    posterior.map { case (l, _) => ev(l).generator }.sample(nSamples)
+
   def animate(sampler: Sampler,
               warmupIterations: Int,
               syntheticSamples: Int,
-              logBins: Int = 3)(implicit rng: RNG): Unit = {
+              logBins: Int = 3,
+              nSamples: Int = 10)(implicit rng: RNG): Unit = {
+    val newGoldset = posteriorSamples(nSamples)
```
avi-stripe: since we have the `Real` version of the posterior (which we use for binning), should we use that also for the goldset? Then our goldset is always a `List[Double]` which simplifies things.

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -0,0 +1,48 @@
+package com.stripe.rainier.core
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.sampler._
+
+trait SBCModel[L, T] {
+  implicit val rng: RNG = ScalaRNG(1528673302081L)
+  def sbc: SBC[L, T]
+  def main(args: Array[String]): Unit = sbc.animate(HMC(1), 10000, 1000)
```
avi-stripe: we probably want the implementations of this trait to be able to override these sampling params

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -0,0 +1,48 @@
+package com.stripe.rainier.core
```
avi-stripe: have you tried this in the test package yet?

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.sampler._
+
+trait SBCModel[L, T] {
```
avi-stripe: could drop the type params if we moved to goldset always being Double

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -0,0 +1,48 @@
...
+  def goldset: List[Double] =
+    List(1.9444409266417129, 4.884477223783541, 3.166384786757497,
+      1.340419862750101, 62.45333008770111, 21.260796478808622,
+      2.145416458600115, 0.5978183569453146, 2.745814563529916,
+      2.326159027201714)
+  def description = "LogNormal(x,x) with LogNormal(0,1) prior"
+}
+
+/**
+  * Note: this are made-up goldsets. SBC on these is wildly slow.
```
avi-stripe: Hmmm we should look into that :)

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -0,0 +1,48 @@
...
+}
+
+object SBCUniformNormal extends SBCModel[Continuous, Double] {
+  def sbc = SBC(Uniform(0, 1))((x: Real) => Normal(x, 1))
+  def goldset =
+    List(0.770217434378013, 1.2877991075049935, 1.0761575302777342,
+      0.06063426627950119, 2.235772959187611, 1.5916139704386643,
+      0.4680530806057483, -1.3360844052727425, 0.6558428534558145,
+      0.4233930395976211)
+  def description = "Normal(x,1) with Uniform(0,1) prior"
```
avi-stripe: everywhere we have `def` here we could use `val`

rainier-tests/src/test/scala/com/stripe/rainier/core/SBCTest.scala
```diff
@@ -0,0 +1,17 @@
+package com.stripe.rainier.core
+
+import org.scalatest.FunSuite
+
+class SBCTest extends FunSuite {
+
+  def check[L, T](sbcModel: SBCModel[L, T]): Unit = {
```
avi-stripe: Love it.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
```diff
@@ -35,6 +40,10 @@ final case class SBC[L, T](priorGenerators: Seq[Generator[Double]],
       val timeRemaining = timeTaken * (reps - i) / i
       plot(list, bins, i, reps, lower, upper, timeRemaining)
     }
+    println(s"\nnew goldset:")
```
avi-stripe: I might move this into the `main` since we might be using SBC not from SBCModel

avi-stripe (changes requested):

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: I always get a `[error] java.lang.RuntimeException: No main class detected.` when I put it test....by which I mean, putting it into `rainier-tests/src/test/scala/com/stripe/rainier/core` and leaving this is `package com.stripe.rainier.core`, then trying to run it with
```
sbt:rainier-tests> run
[info] Including from cache: asm-tree-6.0.jar
[info] Including from cache: asm-6.0.jar
[info] Checking every *.class/*.jar file's SHA-1.
[info] Merging files...
[warn] Merging 'module-info.class' with strategy 'discard'
[warn] Strategy 'discard' was applied to a file
[info] Assembly up to date: /Users/mio/stripe/rainier/.asm-deps/target/asmDeps-assembly-0.1.0.jar
[info] Packaging /Users/mio/stripe/rainier/rainier-core/target/scala-2.12/rainier-core_2.12-0.1.0.jar ...
[info] Done packaging.
[error] java.lang.RuntimeException: No main class detected.
```

mio-stripe:
Thanks for the comments! I'll on them more tomorrow.

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
avi-stripe: @mio-stripe try putting in `rainier-tests/src/main/scala/com/stripe/rainier/core` (note the `src/main`)

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: 👍

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: 🎉 thanks! Definitely should be in `src/main` :)

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: I changed everything to a `val` and got 

```
[info] Running com.stripe.rainier.core.SBCUniformNormal
[error] (run-main-1) java.lang.ExceptionInInitializerError
[error] java.lang.ExceptionInInitializerError
[error] 	at com.stripe.rainier.core.SBCUniformNormal.main(SBCModel.scala)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[error] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[error] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[error] 	at java.lang.reflect.Method.invoke(Method.java:498)
[error] 	at sbt.Run.invokeMain(Run.scala:93)
[error] 	at sbt.Run.run0(Run.scala:87)
[error] 	at sbt.Run.execute$1(Run.scala:65)
[error] 	at sbt.Run.$anonfun$run$4(Run.scala:77)
[error] 	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[error] 	at sbt.util.InterfaceUtil$$anon$1.get(InterfaceUtil.scala:10)
[error] 	at sbt.TrapExit$App.run(TrapExit.scala:252)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] Caused by: scala.UninitializedFieldError: Uninitialized field: /Users/mio/stripe/rainier/rainier-tests/src/main/scala/com/stripe/rainier/core/SBCModel.scala: 29
```

then changed two things back and it worked. Seems like the things that are not implemented in the trait have to be defs and the rest can be vals.

avi-stripe: Ah, this is about initialization order - you need to make sure the vals are in dependency order from top to bottom (or make them `def` or `lazy val`).

mio-stripe:
I addressed most comments. I'll work on removing the type parameters from SBCModel this afternoon.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
mio-stripe: 👍 I switched to use `RandomVariable.record` on the real part of the posterior to get a `List[Double]`.

mio-stripe:
Hmm, okay, I'm running into some type constraint problems in trying to remove the type parameters from `SBCModel` now that `SBC` has a type constraint on `L` rather than implicit evidence. I'm pretty sure there should be a way to do this. Here's an example. 

```
scala> case class Foo[T, L <: Distribution[T]](l: L)
defined class Foo

scala> trait Yo { def foo[T, L <: Distribution[T]]: Foo[T, L] }
defined trait Yo

scala> object BYo extends Yo { def foo[Double, Continuous] = Foo[Double, Continuous](LogNormal(0,1)) }
<console>:26: error: type arguments [Double,Continuous] do not conform to method apply's type parameter bounds [T,L <: com.stripe.rainier.core.Distribution[T]]
       object BYo extends Yo { def foo[Double, Continuous] = Foo[Double, Continuous](LogNormal(0,1)) }
                                                                ^
<console>:26: warning: type parameter Double defined in method foo shadows class Double defined in package scala. You may want to rename your type parameter, or possibly remove it.
       object BYo extends Yo { def foo[Double, Continuous] = Foo[Double, Continuous](LogNormal(0,1)) }
                                       ^
<console>:26: warning: type parameter Continuous defined in method foo shadows trait Continuous defined in package core. You may want to rename your type parameter, or possibly remove it.
       object BYo extends Yo { def foo[Double, Continuous] = Foo[Double, Continuous](LogNormal(0,1)) }
```

and

```
scala> object BYo extends Yo { def foo = Foo[Double, Continuous](LogNormal(0,1)) }
<console>:26: error: object creation impossible, since method foo in trait Yo of type [T, L <: com.stripe.rainier.core.Distribution[T]]=> Foo[T,L] is not defined
       object BYo extends Yo { def foo = Foo[Double, Continuous](LogNormal(0,1)) }
```

avi-stripe:
@mio-stripe why does `SBCModel` need type params at all? Can't it have none and its `sbc` just be `SBC[_,_]`?

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: Still working on this.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
mio-stripe: Done

mio-stripe:
Yep, that did it. Removing them on the trait but having `def sbc[T, L]` definitely did not work.

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: Done

rainier-core/src/main/scala/com/stripe/rainier/core/SBCModel.scala
mio-stripe: Yep! For now I left these SBCModels in, but removed them from SBCTest so that tests pass.

avi-stripe (approved):

mio-stripe:
@avi-stripe should we hold off on merging this since it has the listmap change or forge ahead?

avi-stripe:
@mio-stripe ah good point. Let's hold off and keep this on the to-be-merged queue.

mio-stripe:
Listmap https://github.com/stripe/rainier/pull/182 is now the base branch and this contains only SBC-related changes.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
```diff
@@ -19,6 +19,15 @@ final case class SBC[T, L <: Distribution[T]](priors: Seq[Continuous],
 
   val priorGenerator = Generator.traverse(priors.map(_.generator))
 
+  def posteriorSamples(nSamples: Int)(implicit rng: RNG): List[T] =
```
avibryant: Just realized that this is only exercising the logDensity of the prior, and the generator of the likelihood. I think we should be trying to exercise both functions for both distributions. I think we could do that with something like `model(1000).sample(nSamples)` here.

mio-stripe: Ah, good point. I'll make this change in https://github.com/stripe/rainier/pull/205 so I don't have to do any more git coordination.

PR #201, opened 2018-09-10
optimizations for simple sample generation

avi-stripe:
This adds a couple of simple optimizations that can dramatically speed up sampling from `Generator`s derived from simple distributions:

* in `Evaluator`, bypass the cache for `Constant` since we can just read the value directly from the object
* in `Injection` add an optional `fastForwards` version of the `forwards` method that works in `Double`-space instead of `Real`-space to avoid round-tripping through `Real` where that's not needed.

These were motivated by work Jack D is doing that relies on forwards-only generative use of Rainier.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Evaluator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Evaluator.scala
@@ -2,13 +2,17 @@ package com.stripe.rainier.compute
 
 class Evaluator(var cache: Map[Real, Double]) extends Numeric[Real] {
 
-  def toDouble(x: Real): Double = cache.get(x) match {
-    case Some(v) => v
-    case None => {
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Evaluator.scala
```diff
@@ -2,13 +2,17 @@ package com.stripe.rainier.compute
...
 
-  def toDouble(x: Real): Double = cache.get(x) match {
-    case Some(v) => v
-    case None => {
-      val v = eval(x)
-      cache += x -> v
-      v
-    }
+  def toDouble(x: Real): Double = x match {
+    case Constant(v) => v.toDouble
```
mio-stripe, line 6: Niiice

mio-stripe:
Very nice!
LGTM

PR #203, opened 2018-09-12
[WIP] Kahan summation for numerical stability

mio-stripe:
This adds Kahan summation for Array[Double].


mio-stripe:
@avi-stripe I'd like to test more Arrays. Is the model you mentioned that built a big Array of Doubles that worked one way but not another still floating around?

avibryant:
@mio-stripe this was the change I was mentioning to you earlier: https://github.com/stripe/rainier/pull/119 .
Not sure if you can easily extract a test case from that.

avibryant:
What would be especially cool in these tests is some kind of demonstration that normal floating point addition would *not* come up with the right answer (in the cases where that's true).

mio-stripe:
Okay, I rewrote the tests in scalacheck so they are a bit more general. One shows that Kahan summation gets the expected answer and is as good or better than the naive sum and the other shows that the naive sum is just wrong where Kahan-Neumaier summation is correct. Not really clear that we need both variants, Kahan and Neumaier, but I wanted to make sure both were right first. I don't quite get how to make a test case out of https://github.com/stripe/rainier/pull/119 but since it's related to ordering, I wonder what effect the listmap change https://github.com/stripe/rainier/pull/182 will have on it.

mio-stripe:
I pulled this code into https://github.com/stripe/rainier/pull/262
Not sure if we need this vanilla version for Double, but it's good to have as a reference for the moment.

PR #204, opened 2018-09-12
batched target and context

avibryant:
This enables batched evaluation of log-density for simple non-gradient cases like `Walkers`. The major pieces of functionality are:

* `Target.batched(k)`. Given a `Target` with a `placeholders: Map[Variable, Array[Double]]`, we assume that the `Array[Double]` values all have the same length, `n`. If we divide each of these into `k` batches, we end up with `m = n/k` items per batch. We can represent this by a dense `k x m` matrix, with a "remainder" of `n - (k*m)` items that don't fit inside it. We want to ultimately do one evaluation for each row of this matrix (ie each batch); to achieve this, we copy each `Variable` in the original `placeholders` map `m` times, and produce a new copy of the target's `real` that has these cloned `Variable`s substituted for the original ones, and then create a new `Target` whose `real` is the sum of those copies, and whose `placeholders` has the new `Variable` objects and the appropriate matrix columns as its arrays. At the same time, we inline the `remainder` as constants into `real` to produce a fully inlined (no placeholders) version that covers those items.
* `Context.apply(k,targets)`. Given a list of targets and a number of batches, this uses `batched` on all of them to produce a new target with the correct number of batches, as well as a remainder. It can then sum all of the remainders, and, because all of the `Targets` have been normalized to have the same number of items in their placeholder arrays, it can union all of them also, leading to a `Context(base: Real, batched: Target)`.
* `Context.variables` has been modified to look at both the `base` and the `real` inside the batched target, not *not* to consider the variables in the target's placeholders (since we don't want them in the gradient).
* `compileBatched` produces an `Array[Double] => Double` function for the target whose input is just the parameters array (ie the same as `variables`), but that internally compiles a function that takes the `placeholders` as input also, and then divides up the placeholders into batches and evaluates that compiled function once per batch, summing the results and returning that. (This summation is one place we could use kahan summation btw). It's worth noting some potential inefficiency here in that any intermediate values in that compiled function that do *not* depend on the placeholder variables still get recomputed for each batch; a smarter implementation of this which coordinated more deeply with the compiler could avoid that.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
@@ -7,7 +7,7 @@ case class Context(base: Real, batched: Target) {
     val allVariables =
       RealOps.variables(base) ++
         RealOps.variables(batched.real) --
-        batched.placeholders.keySet
+        batched.placeholderVariables
     allVariables.toList.sortBy(_.param.sym.id)
   }
 
@@ -24,20 +24,19 @@ case class Context(base: Real, batched: Target) {
 
   //not terribly efficient
   private def compileBatched: Array[Double] => Double = {
-    if (batched.placeholders.isEmpty) {
+    if (batched.nRows == 0) {
       compiler.compile(variables, base + batched.real)
     } else {
-      val placeholders = batched.placeholders.keys.toList
-      val cf = compiler.compile(variables ++ placeholders, batched.real)
+      val cf = compiler.compile(variables ++ batched.placeholderVariables,
+                                batched.real)
 
-      val nBatches = batched.placeholders.head._2.size
       val result = { input: Array[Double] =>
-        0.until(nBatches)
+        0.until(batched.nRows)
           .map { i =>
             val extraInputs =
-              placeholders.toArray.map { v =>
+              batched.placeholderVariables.map { v =>
                 batched.placeholders(v)(i)
-              }
+              }.toArray
             cf(input ++ extraInputs)
           }
           .sum
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
@@ -1,10 +1,71 @@
 package com.stripe.rainier.compute
 
-case class Context(density: Real) {
+case class Context(base: Real, batched: Target) {
   val compiler: Compiler = IRCompiler(200, 100)
-  val variables: List[Variable] = RealOps.variables(density).toList
-  lazy val gradient: List[Real] = Gradient.derive(variables, density).toList
 
-  def compileDensity: Array[Double] => Double =
-    compiler.compile(variables, density)
+  val variables: List[Variable] = {
+    val allVariables =
+      RealOps.variables(base) ++
+        RealOps.variables(batched.real) --
+        batched.placeholders.keySet
+    allVariables.toList.sortBy(_.param.sym.id)
+  }
+
+  //convenience method for simple cases like Walkers
+  def compileDensity: Array[Double] => Double = {
+    val baseCF = compiler.compile(variables, base)
+    val batchCF = compileBatched
+
+    val result = { input: Array[Double] =>
+      baseCF(input) + batchCF(input)
+    }
+    result
+  }
+
+  //not terribly efficient
+  private def compileBatched: Array[Double] => Double = {
+    if (batched.placeholders.isEmpty) {
+      compiler.compile(variables, base)
+    } else {
+      val placeholders = batched.placeholders.keys.toList
+      val cf = compiler.compile(variables ++ placeholders, batched.real)
+
+      val nBatches = batched.placeholders.head._2.size
+      val result = { input: Array[Double] =>
+        0.until(nBatches)
+          .map { i =>
+            val extraInputs =
+              placeholders.toArray.map { v =>
+                batched.placeholders(v)(i)
+              }
+            cf(input ++ extraInputs)
+          }
+          .sum
+      }
+      result
+    }
+  }
+
+  //these are for backwards compatibility to allow HMC to keep working
+  //they will not make sense in the general case going forward
+  lazy val density: Real = base + batched.inlined
+
+  lazy val gradient: List[Real] =
+    Gradient.derive(variables, density)
+}
+
+object Context {
+  def apply(real: Real): Context = Context(real, Target.empty)
+  def apply(nBatches: Int, targets: Iterable[Target]): Context = {
+    val (base, batched, batches) =
+      targets
+        .foldLeft((Real.zero, Real.zero, Map.empty[Variable, Array[Double]])) {
+          case ((oldBase, oldBatched, oldBatches), target) =>
+            val (batchedTarget, remainder) = target.batched(nBatches)
+            (oldBase + remainder,
+             oldBatched + batchedTarget.real,
+             oldBatches ++ batchedTarget.placeholders)
+        }
+    Context(base, new Target(batched, batches))
+  }
 }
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -18,6 +18,45 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
...
+
+  def batched(nBatches: Int): (Target, Real) =
+    if (placeholders.isEmpty)
+      (this, Real.zero)
+    else if (nBatches == 1)
+      (Target.empty, inlined)
+    else {
+      val n = nRows(placeholders)
+      val variables = placeholders.keys.toList
+      val nColumns = n / nBatches
```
mio-stripe: Okay, I think I get what we're doing, but I had to think more even after we talked it out. Let me know if this sounds right. By "an `n x m` matrix" I mean "a matrix with `n` rows and `m` columns."
So, placeholders is an `n x v` matrix with `n` rows, the lengths of the arrays and `v` columns, the number of variables. If `nBatches = b` and `m = n / b`, ignoring remainders for the moment, we are going to rearrange this into a `b x vm` matrix where we'll think of each row of this new matrix as a batch. We'll do this as follows: take our original `n x v` matrix and slice it horizontally into `m` slices each of which has `b` rows. This is already potentially confusing because the number of slices here is _not_ `nBatches = b`, it is `m`. Our slices are still stacked vertically, one on top of another. Now, we take these slices and place them side by side, left to right. By placing our `m` slices, each of dimension `b x v` next to each other, we get the `b x vm` matrix we want.

Let's do an example: suppose that we have a `6 x 2` matrix with entries `a_{ij}` (i = row, j = column) and we want `b = 2` batches. Then `m = 3`. Rearranging the slices, we get the matrix

```
a11  a12  a31  a32  a51  a52
a21  a22  a41  a42  a61  a62 
```

The first slice contains the first two rows and all of the columns, the next slice the next two rows and all of the columns, and so on. Here, `copies` makes copies of all of the variables and does so `m` times, one for each slice. To make the new placeholders, we iterate over the slices and take the coefficients from the original matrix which are in the slice we're in.

My only code comment so far is: I see that `n / nBatches` is the number of slice we are arranging side-by-side, but calling it `nColumsn` is pretty confusing. I think `nSlices` would be better.

Apologies for so much talking with so few helpful suggestions. I'll have more after lunch.

avibryant: Yes, that's correct. I agree that the naming could be better, I'll work on that before we merge. (FWIW, the idea behind "nColumns" was something more like "ihow many columns do we end up with *for each v*")

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -18,6 +18,45 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
...
+      val copies = 0.until(nColumns).map { i =>
+        val columnVariables = variables.map { v =>
+          v -> new Variable
+        }.toMap
+        val columnReal = PartialEvaluator.inline(real, columnVariables)
+        (i, columnReal, columnVariables)
+      }
+      val newReal = Real.sum(copies.map(_._2))
```
mio-stripe: I understand the "make a copy of the original `real` with the cloned variables substituted for the original ones" but don't totally get the "and them sum these to make the `newReal`." I get that we want a single `newReal` at the end and don't _really_ see another way to make one, but why is the sum what we want?

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -18,6 +18,45 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
...
+  def batched(nBatches: Int): (Target, Real) =
+    if (placeholders.isEmpty)
+      (this, Real.zero)
+    else if (nBatches == 1)
+      (Target.empty, inlined)
+    else {
+      val n = nRows(placeholders)
+      val variables = placeholders.keys.toList
+      val nColumns = n / nBatches
+      val copies = 0.until(nColumns).map { i =>
```
mio-stripe: Since `batched` already has dense logic, even though the code is not particularly dense, it might still help to move the `copies, newPlaceholders` and `remainder` logic into private helper functions so the body of `batched` itself is super streamlined.

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -18,6 +18,45 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
...
+        val columnVariables = variables.map { v =>
+          v -> new Variable
+        }.toMap
+        val columnReal = PartialEvaluator.inline(real, columnVariables)
+        (i, columnReal, columnVariables)
+      }
+      val newReal = Real.sum(copies.map(_._2))
+      val newPlaceholders = copies.flatMap {
```
mio-stripe: This `flatmap...map...` could be a `for` which would remove a level of nesting.

rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
```diff
@@ -1,10 +1,71 @@
...
-  val variables: List[Variable] = RealOps.variables(density).toList
-  lazy val gradient: List[Real] = Gradient.derive(variables, density).toList
 
-  def compileDensity: Array[Double] => Double =
-    compiler.compile(variables, density)
+  val variables: List[Variable] = {
+    val allVariables =
+      RealOps.variables(base) ++
+        RealOps.variables(batched.real) --
+        batched.placeholders.keySet
```
mio-stripe: I don't totally understand this subtraction: when we make a batched Target, we make `m` copies of all of the variables in the original placeholders, make copies of the original real where we substitute in the cloned variables, then sum these to make the `newReal` so the variables in `batched.real` look like they are all of the copies we make of all of the original variables. Those are also all of the variables in the `newPlaceholders` so it looks to me like `RealOps.variables(batched.real)` and `batched.placeholders.keySet` are the same. What am I missing?

rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
```diff
@@ -1,10 +1,71 @@
...
+
+  //not terribly efficient
+  private def compileBatched: Array[Double] => Double = {
+    if (batched.placeholders.isEmpty) {
+      compiler.compile(variables, base)
+    } else {
+      val placeholders = batched.placeholders.keys.toList
+      val cf = compiler.compile(variables ++ placeholders, batched.real)
+
+      val nBatches = batched.placeholders.head._2.size
```
mio-stripe: If `nRows` were a val on Target a la `Target(real, placeholders) {  def nRows = placeholders.head._2.size }` then this line could be `val nBatches = batched.nRows` which would be another indication that batching rearranges a Target to have `nBatches` rows.

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
avibryant: That's just the semantics of `Target`. We use it when we have a log-likelihood function `f` and a list of observations `x`, where the non-batchable version would be `x.map(f).sum`.

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
avibryant: ok sounds good

rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
avibryant: The original `real` in the `Target`, and hence the `newReal` in the batched `Target`, can also refer to `variables` that are not placeholders (ie, that are true parameters). These are the ones we want (because they're the ones we care about the gradients for).

rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
avibryant: Good idea, thanks.

rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
mio-stripe: aaah, got it.

mio-stripe:
Nice!
LGTM

PR #205, opened 2018-09-18
Move more tests over to SBC

mio-stripe:
Make some more SBC versions of tests.


rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -67,7 +67,7 @@ final case class Geometric(p: Real) extends Discrete {
   * @param n Total number of failures
   * @param p Probability of success
   */
-final case class NegativeBinomial(n: Real, p: Real) extends Discrete {
+final case class NegativeBinomial(p: Real, n: Real) extends Discrete {
```
mio-stripe, line 71: I switched the order so that it is the same as `Binomial` (even though I think both should be the other way :shrug:).

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
```diff
@@ -13,8 +13,8 @@ final case class SBC[T, L <: Distribution[T]](
 
   val priorGenerator = Generator.traverse(priorGenerators)
 
-  def posteriorSamples(nSamples: Int)(implicit rng: RNG): List[Double] =
-    posterior.map { case (_, r) => r }.record(nSamples).params.flatten
+  def posteriorSamples(nSamples: Int)(implicit rng: RNG): List[T] =
+    posterior.map { case (l, _) => l.generator }.sample(nSamples)
```
mio-stripe: This is what we had originally and we changed it so that it would always return a `List[Double]`. The problem is that that `List[Double]` consists of values of the prior parameters that we then feed into the model again in `replay` to get a sample from the posterior. For SBC, we actually want a sample from the posterior or else different posteriors with the same priors will have the same values here and checking goldsets won't really tell us much.

rainier-tests/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -60,3 +61,67 @@ object SBCLaplace extends SBCModel {
...
+  def goldset =
+    List(22, 12, 28, 21, 19, 5, 14, 7, 3, 4, 7, 8, 26, 8, 2, 3, 9, 6, 5, 11, 9)
+  val description = "NegativeBinomial(x, 10) with Uniform(0,1) prior"
+}
+
+/**
+ * sbc.animate throws an error for both of these approximations.
+ * Generating the posterior samples works fine.
+ */
+object SBBinomialPoissonApproximation extends SBCModel {
```
mio-stripe: SBC.animate throws errors for both of these approximations. Still debugging.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
mio-stripe: Two things actually: `posterior.map { case (_, r) => r }.record(nSamples).params.flatten` is _definitely_ not what we want because `r` is literally just the prior again, but even `posterior.map { case (l, _) => l }.record(nSamples).params.flatten` is not what we want because it _also_ returns the prior, this time because of how `record` works.

mio-stripe:
For a reason that I don't yet understand, `SBCBinomial` runs fine, but the two `SBCBinomial...Approximation`s immediately fail in the `SBC.sample` step in `SBC.animate` with `java.lang.RuntimeException: No value provided for com.stripe.rainier.compute.Variable@stuff`. A minimal failing example that reproduces the `SBC.sample` logic is 

```
import com.stripe.rainier.core._
import com.stripe.rainier.compute._

implicit val rng: RNG = ScalaRNG(1528673302081L)
val emptyEvaluator = new Evaluator(Map.empty)
val prior = Uniform(0,1)
val priorParams = prior.param.value
val priorGenerator = prior.generator
val posterior = prior.param.map{ x => Normal(x,10) }
val trueValues = priorGenerator.get(rng, emptyEvaluator)
implicit val trueEval = new Evaluator(Map(priorParams -> trueValues))
lazy val syntheticValues = posterior.map{ c => c.generator.repeat(10) }.get
```

I'm a little confused that this example fails because I think it is doing the same thing as `SBCBinomial` which passes, but it does fail in the same way as the other failing SBCModels.

I tried changing `syntheticValues` in `SBC.sample` from 

```posterior.map { case (l, _) => l.generator.repeat(syntheticSamples) }.get```

to 

```posterior.map { case (l, _) => l.generator }.sample(syntheticSamples)```

which ran, but clearly broke things because the SBC histograms were totally skewed. This makes me think I'm now doing it the wrong way in `posteriorSamples`, but I don't totally get the difference between these two.

**Comment by mio-stripe** [comment]
Also, pretty confused by this build failure: it is caused by SBCTest failing, but it passes on my machine.

**Comment by mio-stripe** [comment]
I see the problem: calling `param` on a `Continuous` creates a new `Variable` and in my example, I call it twice, once in `priorParams` and once in `posterior`. This doesn't happen in `SBCBinomial`, but it does in the `SBCBinomialApproximation`s _I think_ because those are `Uniform(a,b)` for `a` and `b` not 0 and 1. Heading out in a few, but I'll trace through why that is happening for non-standard Uniforms and think about a solution.

**Comment by mio-stripe** [comment]
The initial SBC branch https://github.com/stripe/rainier/pull/200 is now the base branch so this branch will only contain the SBCModels added on top of that.

**Comment by avibryant** [comment]
Thanks for all the cleanup work stacking the PRs. LGTM

File: rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala [hunk 4]
```diff
@@ -22,9 +22,11 @@ final case class SBC[T, L <: Distribution[T]](priors: Seq[Continuous],
   def posteriorSamples(nSamples: Int)(implicit rng: RNG): List[T] =
     RandomVariable
       .traverse(priors.map(_.param))
-      .map { priorParams =>
+      .flatMap { priorParams =>
```
mio-stripe: @avi-stripe addressing your comment about exercising the logDensity and generator for both the prior and likelihood: I'm almost doing `model(1000).sample(nSamples)`, but `SBC.fit` returns the prior so I did this to get samples from the posterior. It looks like we've got everything now because we use both generators in `synthesize` (and the posterior generator again here) and flatmapping with a call to `RandomVariable.fit` uses both logDensities. This is pretty similar to `fit` and `synthesize` so lemme know if you see a better way to do this.

mio-stripe:
Welp, something's not right because there's some non-determinism in the posteriorSamples now.

mio-stripe:
The SBC tests seem to pass often, but not always.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
avibryant: why do we care, though? can't we just use the (inferred) prior parameter as our output?

avibryant:
@mio-stripe as in we have non-deterministic non-determinism? Oh great.

rainier-core/src/main/scala/com/stripe/rainier/core/SBC.scala
mio-stripe: Oh, I see, since we're fitting now, that incorporates information about the likelihood so we'll get different prior params even for models with the same prior distribution. Yeah, that's a good point. I'll try switching it.

mio-stripe:
Using `model(1000).sample(nSamples)` where we get the inferred prior params after fitting, it is _deterministically_ non-deterministic: we get close, but different results each time that usually agree on the first few decimal places, but not all the way. I'm guessing this is an evaluation order issue, but the way I handled that before was to make `SBCModel.samples` a `val` so that it was computed on initialization, before `SBC.animate` ran. I don't yet get why this is happening now. I think the non-deterministic non-determinism before was a symptom of the same problem.

mio-stripe:
Okay, heading out, but I'll try some variations on the plane tomorrow and see if I can figure it out. There's a definitely a diagram that doesn't quite commute.

avibryant:
@mio-stripe what happens if instead of using an implicit `RNG`, you explicitly pass in a newly created RNG with a fixed seed? I'd be surprised if we still saw any ordering issues after that.

avibryant:
edit: pass it in to posteriorSamples, I mean

mio-stripe:
So sticking the rng inside of `posteriorSamples` didn't solve the problem and I found a few things:
* running sbc.animate with the same rng produces different histograms
* sbc.synthesize _is_ deterministic.
* sbc.fit(values).sample(nSamples) is _not_ deterministic

So the plot thickens.

mio-stripe:
Coming back to this, the histograms look bad and different for different runs with Walkers and good and different for different runs with HMC.
I switched `posteriorSamples` back to be just `sbc.fit(values).sample(nSamples) ` and that seems to work fine using HMC for the SBCModels with Continuous posteriors—the tests consistently pass—but not for the Discrete ones e.g.

```
scala> def sbc = SBC[Int, Discrete](Uniform(0, 1))((x: Real) => Bernoulli(x))
scala> implicit val rng: RNG = ScalaRNG(1528666602081L)
scala> val values = sbc.synthesize(1000)._1
scala> implicit val rng: RNG = ScalaRNG(1528666602081L)
scala> sbc.fit(values).sample(20)
res6: List[Double] = List(0.36229695299084447, 0.36229695299084447, 0.36229695299084447, 0.36229695299084447, 0.36229695299084447, 0.36937742763054454, 0.3678052121475404, 0.3698669740622792, 0.3698669740622792, 0.3698669740622792, 0.3698669740622792, 0.3557462709839959, 0.3862766900922043, 0.39735959642099467, 0.36738111931895373, 0.3498565819134614, 0.3812066154231802, 0.3596768241441413, 0.3596768241441413, 0.38236636674933344)
scala> implicit val rng: RNG = ScalaRNG(1528666602081L)
scala> sbc.fit(values).sample(20)
res7: List[Double] = List(0.362296835357449, 0.362296835357449, 0.3429838849648832, 0.3429838849648832, 0.3979500855731633, 0.36910200456218867, 0.36797970957003595, 0.37009554222613894, 0.37009554222613894, 0.37009554222613894, 0.37009554222613894, 0.3560223146935066, 0.3870277337819613, 0.39589845559192827, 0.36585177109732536, 0.3503387602869432, 0.3823995426585505, 0.35923105543951395, 0.35923105543951395, 0.3830161302477496)
```

Using Walkers for the Discrete ones does produce the same posterior samples
```
scala> sbc.fit(values).sample(Walkers(4), 100000, 20)
res0: List[Double] = List(0.36715818439808845, 0.3683298414003457, 0.36284446275987176, 0.3732451892218014, 0.3667256741281268, 0.3639690416072447, 0.36312003216062594, 0.37523573847491937, 0.36610541470879615, 0.3642600266812339, 0.362651301173622, 0.3719510738157602, 0.367457819663703, 0.3636338866876684, 0.36225408389684394, 0.37270186702317276, 0.36524380567075143, 0.36290022896459495, 0.3617773479346795, 0.36987639611425965)

scala> implicit val rng: RNG = ScalaRNG(1528666602081L)
rng: com.stripe.rainier.sampler.RNG = ScalaRNG(1528666602081)

scala> sbc.fit(values).sample(Walkers(4), 100000, 20)
res1: List[Double] = List(0.36715818439808845, 0.3683298414003457, 0.36284446275987176, 0.3732451892218014, 0.3667256741281268, 0.3639690416072447, 0.36312003216062594, 0.37523573847491937, 0.36610541470879615, 0.3642600266812339, 0.362651301173622, 0.3719510738157602, 0.367457819663703, 0.3636338866876684, 0.36225408389684394, 0.37270186702317276, 0.36524380567075143, 0.36290022896459495, 0.3617773479346795, 0.36987639611425965)
```

So, dunno. Let's revisit after we merge other recent changes. Seems like we could probably solve the histogram non-determinacy and posteriorSamples non-determinacy separately.

mio-stripe:
The problem was that HMC would hang in `findReasonableStepSize` causing SBC to hang. This is a transient problem that we should fix, but rerunning SBC when that doesn't happen, the histograms look maybe officially okay but wonky? Large bars at the ends e.g.

```
[info] Running com.stripe.rainier.core.SBCBinomialPoissonApproximation

Running simulation-based calibration.
Repetition 320/320. Estimated time remaining: 0:00:00
Samples/sec: 43. Max rHat: 274.320. Max thinning factor: 199
99% of bins should end up between the [ and ] quantile markers

##########################[##############################]##########
##########################[##################            ]
######################### [                              ]
##########################[###########                   ]
##########################[                              ]
##########################[#########                     ]
##########################[###                           ]
##########################[##############################]##
```

mio-stripe:
@avi-stripe 🎉 all of the tests pass and the histograms look good! I actually think it's better for the histograms _not_ to be deterministic so that you can see what natural variation of histograms looks like. The posteriorSamples are now which is obviously the crucial thing.

PR #208, opened 2018-09-21
fix sbc for more complex cases

avibryant:
(this is a standalone version of a fix I already added to mio's branch)

This gives the `SBC` case class the `(priors: Seq[Continuous], fn: Seq[Real] => (L, Real))` interface that used to be just a convenience `apply` method on its companion object. By giving it the `prior => noise dist` as a function, rather than a pre-constructed RV, we can evaluate it in double-space rather than Real-space when we're synthesizing data, which is faster, simpler, and more robust. The parallel structure between the synthesis mode and the inference mode is now also more obvious.


mio-stripe:
LGTM

PR #209, opened 2018-09-24
sbc-based benchmarks

avibryant:
Consistent with the SBC-based testing, start to move to benchmarks that also use the SBC objects as a model specification - this does away with any need to explicitly provide synthetic data as part of a benchmark.

This does not use the same SBCModel objects as the tests do, because I'm not sure how well the benchmark annotations would mesh with the test classes, and because it's not clear you want to benchmark the exact same models as you use for testing. But the basic idea is the same.

This also allows us to benchmark batched/placeholdered vs inlined treatment of data, though I'm not happy with the need to define separate classes for different parameterizations of the batching and data size; I'd rather automatically do a kind of grid search over that.

### DIFF
--- a/rainier-benchmark/src/main/scala/com/stripe/rainier/bench/SBCBenchmark.scala
+++ b/rainier-benchmark/src/main/scala/com/stripe/rainier/bench/SBCBenchmark.scala
@@ -1,0 +1,24 @@
+package com.stripe.rainier.bench
+
+import org.openjdk.jmh.annotations._
+import java.util.concurrent.TimeUnit
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.core._
+import com.stripe.rainier.sampler._
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 3, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(3)
+@Threads(4)
+@State(Scope.Benchmark)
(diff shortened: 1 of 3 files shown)


mio-stripe:
Nice LGTM

avibryant:
For reference here's some benchmark output from this PR:

```
build                SBCLaplace                      14400.000
build                SBCLaplace100k                 120000.000
build                SBCNormal                       11400.000
build                SBCNormal100k                  128000.000
compile              SBCLaplace                      64000.000
compile              SBCLaplace100k                  60000.000
compile              SBCNormal                         700.000
compile              SBCNormal100k                   54000.000
compileInlined       SBCLaplace                      27000.000
compileInlined       SBCLaplace100k                8000000.000
compileInlined       SBCNormal                         360.000
compileInlined       SBCNormal100k                     420.000
inline               SBCLaplace                          0.600
inline               SBCLaplace100k                1200000.000
inline               SBCNormal                           1.200
inline               SBCNormal100k                 1120000.000
run                  SBCLaplace                          4.000
run                  SBCLaplace100k                   7000.000
run                  SBCNormal                           0.600
run                  SBCNormal100k                    7200.000
runInlined           SBCLaplace                          4.000
runInlined           SBCLaplace100k                   2400.000
runInlined           SBCNormal                           0.600
runInlined           SBCNormal100k                       0.600
synthesize           SBCLaplace                        780.000
synthesize           SBCLaplace100k                  96000.000
synthesize           SBCNormal                        1140.000
synthesize           SBCNormal100k                  114000.000
```

PR #210, opened 2018-09-25
in sbcbenchmark, use params to grid search across sample and batch sizes

avibryant:
This creates 3x3=9 variations of sample count and batch size for each sbc-based benchmark.

It's WIP because we still need to adapt the benchmark format script to present this matrix in a reasonable way.


mio-stripe:
Neat

avibryant:
@mio-stripe  rebased and no longer wip, take another look

mio-stripe:
LGTM

PR #211, opened 2018-09-26
optimize `product` for `cats.Monad` instance for `Generator`

kailuowang:
also changed the one for `RandomVariable`to point to zip but it's essentially the same implementation.


avibryant:
LGTM! Thanks for the PR!

PR #212, opened 2018-09-26
make it possible to compute individual outputs of a CompiledFunction

avibryant:
The compiler takes multiple `Real` targets as outputs of a compiled function. Previously, although we would generate private JVM methods for computing these individually, the only public entry point was the `apply` method which takes an array to hold all of their outputs and computes each of them in sequence. This change exposes a public dispatch function where you can access an individual output method by index (but keeps the `apply` method for compatibility and convenience). Currently there's no reason you would want to do this, since the later outputs can depend on globals set up by the previous outputs, so it makes no sense to do anything but compute them all in order anyway. However, as we get into more complex batched or multi-level evaluation scenarios, this will become important.


avibryant:
I think the test failure is sbt caching problems.

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -51,7 +51,8 @@ private[cats] class MonadRandomVariable extends Monad[RandomVariable] {
       f: A => RandomVariable[B]): RandomVariable[B] =
     fa.flatMap(f)
 
-  override def product[A, B](fa: RandomVariable[A], fb: RandomVariable[B]): RandomVariable[(A, B)] =
+  override def product[A, B](fa: RandomVariable[A],
+                             fb: RandomVariable[B]): RandomVariable[(A, B)] =
```
avibryant, line 55: spurious formatting change

rainier-core/src/main/scala/com/stripe/rainier/ir/ApplyMethodGenerator.scala
```diff
@@ -5,12 +5,13 @@ final private case class ApplyMethodGenerator(classPrefix: String,
                                               outputMethods: Seq[Int])
     extends MethodGenerator {
   val isStatic: Boolean = false
-  val methodName: String = "apply"
-  val methodDesc: String = "([D[D[D)V"
+  val methodName: String = "output"
```
avibryant: I guess we should rename this class to OutputMethodGenerator

rainier-core/src/main/scala/com/stripe/rainier/ir/ApplyMethodGenerator.scala
```diff
@@ -5,12 +5,13 @@ final private case class ApplyMethodGenerator(classPrefix: String,
...
+  val methodDesc: String = "([D[DI)D"
 
-  outputMethods.zipWithIndex.foreach {
-    case (m, i) =>
-      storeOutput(i) { callExprMethod(m) }
+  loadOutputIndex()
+  tableSwitch(outputMethods) {
+    case Some(m) => callExprMethod(m)
+    case None    => constant(0.0)
```
gkossakowski: isn't reaching this case a logical error and an exception should be thrown as a fail-fast mechanism?

rainier-core/src/main/scala/com/stripe/rainier/ir/ApplyMethodGenerator.scala
gkossakowski: +1

rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
```diff
@@ -24,34 +24,6 @@ trait Compiler {
                     outputs: Seq[Real]): ir.CompiledFunction
 }
 
-final case class InstrumentingCompiler(orig: Compiler, printEvery: Int)
```
gkossakowski, line 27: is this an incidental change to this commit?

rainier-core/src/main/scala/com/stripe/rainier/ir/MethodGenerator.scala
```diff
@@ -129,18 +122,37 @@ private trait MethodGenerator {
     methodNode.visitInsn(POP2)
   }
 
+  def tableSwitch[K](items: Seq[K])(fn: Option[K] => Unit): Unit = {
```
gkossakowski, line 125: neat! i forgot about tableswitch's existence

gkossakowski (commented):
except for small nits, lgtm

rainier-core/src/main/scala/com/stripe/rainier/ir/ApplyMethodGenerator.scala
avibryant: Yes. I'm being lazy here because I haven't looked into how to generate bytecode for an exception yet.

rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
avibryant: Sorta. I would have had to change this class and I don't use it anymore so it was faster to delete it.

avibryant:
@gkossakowski thanks for the review, will address the nits as best I can and get ci passing, then merge.

PR #213, opened 2018-09-27
Use a multi-output CF to compile batched targets

avibryant:
This is an alternative approach to the single-matrix batched evaluation currently in `develop`. The meat of it is in `Context.compileDensity`.

There's two ideas here:

1. Previously, we did a lot of work to assemble a single matrix with a consistent number of rows (`batches`), which covered potentially multiple `Target` objects from different calls to `fit` in the same model. This allowed us to minimize the number of calls to a compiled function as we streamed through the observational data. 

Here, we instead arrange to make exactly one compiled function call for each observation (plus one call to compute the prior). This makes everything much simpler, at the cost of having potentially many more individual calls.

2. Because we are now making a lot more calls, it is much more important that they be able to reuse intermediate results rather than computing everything from scratch each time. To achieve this, we do the following:

- first, we compile everything (the prior and all of the likelihood targets) into a single CompileFunction, which means that they all share the same `globals` array, and if some value is referenced by more than one of them, it will be stored there and reused rather than recomputed. However, due to https://github.com/stripe/rainier/pull/212, this does not stop us from individually computing the outputs, including calling different output functions different numbers of times, as long as we are careful about the order in which we do so.
- second, we create a *second copy* of each likelihood function coming from a `Target` (via `Target.updater`), which replaces the original placeholder variables with a new set of placeholder variables. We sequence these copies after all of the originals in the compilation. This means that any intermediate computations that occur in the likelihood function, that do *not* depend on the placeholder variables, will turn into global state and the second, "updater" copy will have those as global var references; however, anything that *does* depend on the placeholder variables will, because these have been replaced with new ones, still be recomputed.

With this setup, if we first use the original copy of the likelihood function for the first observation from each target, thereby setting up global state, we can then use the "updater" copy for all subsequence observations, and recompute only what is necessary given the new values there.

We make no attempt in this PR to do the equivalent computation for gradients, although it's a fairly straightforward extension of the above.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
@@ -7,7 +7,7 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
     else
       placeholders.head._2.size
 
-  val placeholderVariables = placeholders.keys.toList
+  val placeholderVariables: List[Variable] = placeholders.keys.toList
   val variables: Set[Variable] = RealOps.variables(real) -- placeholderVariables
 
   def inlined: Real =
@@ -22,6 +22,15 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
       }
       Real.sum(inlinedRows.toList)
     }
+
+  def updater: (Real, List[Variable]) = {
+    val newVars = placeholderVariables.map { _ =>
+      new Variable
+    }
+    val newReal =
+      PartialEvaluator.inline(real, placeholderVariables.zip(newVars).toMap)
+    (newReal, newVars)
+  }
 }
 
 object Target {
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
@@ -40,7 +40,6 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
         i += 1
         result match {
           case l: Line if (l.ax.size > MAX_INLINE_TERMS) =>
-            println("!" + l.ax.size)
             return None
           case _ => ()
         }
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
@@ -10,17 +10,42 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
   val placeholderVariables: List[Variable] = placeholders.keys.toList
   val variables: Set[Variable] = RealOps.variables(real) -- placeholderVariables
 
+  private def inlineRow(i: Int): Real = {
+    val row: Map[Variable, Real] = placeholders.map {
+      case (v, a) => v -> Real(a(i))
+    }
+    PartialEvaluator.inline(real, row)
+  }
+
   def inlined: Real =
     if (placeholders.isEmpty)
       real
     else {
-      val inlinedRows = 0.until(nRows).map { i =>
-        val row: Map[Variable, Real] = placeholders.map {
-          case (v, a) => v -> Real(a(i))
+      val inlinedRows =
+        0.until(nRows).map { i =>
+          inlineRow(i)
         }
-        PartialEvaluator.inline(real, row)
-      }
       Real.sum(inlinedRows.toList)
+    }
+
+  val MAX_INLINE_TERMS = 5000
+  def maybeInlined: Option[Real] =
+    if (placeholders.isEmpty)
+      Some(real)
+    else {
+      var result = Real.zero
+      var i = 0
+      while (i < nRows) {
+        result += inlineRow(i)
+        i += 1
+        result match {
+          case l: Line if (l.ax.size > MAX_INLINE_TERMS) =>
+            println("!" + l.ax.size)
+            return None
+          case _ => ()
+        }
+      }
+      Some(result)
     }
 
   def updater: (Real, List[Variable]) = {
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
@@ -1,76 +1,27 @@
 package com.stripe.rainier.compute
 
 class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
-  val nRows = size(placeholders)
-  val placeholderVariables = placeholders.keys.toList
-
-  def inlined = inl(placeholders)
-
-  def batched(nBatches: Int): (Target, Real) =
+  val nRows =
     if (placeholders.isEmpty)
-      (this, Real.zero)
-    else if (nBatches == 1)
-      (Target.empty, inlined)
-    else {
-      val nCopies = nRows / nBatches
-      val (realCopies, variableCopies) = makeCopies(nCopies)
-      val newReal = Real.sum(realCopies)
-      val newPlaceholders = makePlaceholders(variableCopies, nBatches).toMap
-      val newTarget = new Target(newReal, newPlaceholders)
-      val remainder = makeRemainder(nBatches, nCopies)
-      (newTarget, remainder)
-    }
+      0
+    else
+      placeholders.head._2.size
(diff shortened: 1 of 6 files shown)


gkossakowski:
before I dive into the code, let me summarize how I understand the overall shape of the computation. You have three steps:
1. Partial evaluation step that collapses the size of the computation graph (model) by taking advantage of priors
2. something that computes global values that are constant (?) across observations
3. evaluate the remaining parts of the model for each observation

The confusing thing for me is that if something is constant across observations, it should have gotten folded in 1.

avibryant:
@gkossakowski ok, so the overall log probability computation has this form:

```
prior(θ) +
  x1.map{x => likelihood1(x, θ)}.sum +
  x2.map{x => likelihood2(x, θ)}.sum +
.....
```

where x1, x2, etc are vectors of observation data (or matrices, really, since each row can have multiple doubles).

For simplicity let's take the common case where there's only a single matrix of observation data (so x1, and its likelihood function likelihood1).

We set up a compiled function something like this, where `params` is however wide we need to hold θ, plus two additional elements, let's call them a and b, and where likelihood1a and likelihood1b are two identical functions that differ only in what they use as the `x` observation value: one uses `a` and the other uses `b`.

```
def f(params, globals) =
   Array(
      prior(params, globals),
      likelihood1a(params, globals),
      likelihood1b(params, globals)
   )
```

We always evaluate this passing in θ for the first part of the `params` array, and passing in the same mutable `globals` over and over which will build up state.

The first time we call it we don't care about a or b because we just want to compute the prior.
The second time, we set a=x1[0], our first observation, and get the value of the second output element (likelihood1a).
Then we evaluate it k additional times with b=x1[i], for each of the rest of our observations.

The point is that any intermediate computation that depends solely on θ will get stored in globals in the course of those first two evaluations, and so won't be recomputed by any of the likelihood1b runs. However, state that depends on `a` will not be relied on by it (because it's looking at `b`), and state that relies on `b` will get overwritten and recomputed on each call.

This is all quite fragile and hard to reason about! There are probably layers of abstraction that would make this easier. But it's not clear to me yet what those might be.

gkossakowski:
Thanks for taking time to write this down. Follow up questions:
1. The `x1[0]` is not any special case from the point of view of what likelihood is computing, it's just called to fill in slots in `globals` that depend on only `θ`. You could use any other element of `x1`, right?
2. How exactly the logic inside of likelihood1{a,b} figures out what to recompute? Is there some conditional check inside of these compiled functions? If so, what does it check? This must be related to being able to selectively call pieces of the output function that the previous PR enabled. But I don't see how varying a slot in an array (`a` or `b`) on its own would introduce value reuse.

avibryant:
@gkossakowski 

1. yes, that's right.
2. let me try to make this concrete with an example. Let's say there is only one element in θ, so that's `params[0]`, and only one column in the `x` matrix, so that what I was calling `a` is `params[1]`, and `b` is `params[2]`.

Let's say that we've compiled the following pair of functions (shown as scala though they'd actually be generated bytecode), at first without using the globals array:

```scala
def prior(params: Array[Double], globals: Array[Double]): Double = 
   Math.log(params(0)) * 2

def likelihood(params: Array[Double], globals: Array[Double]): Double = 
   (Math.log(params(0)) + params(0)) * params(1)
```

However, the Compiler, knowing/assuming that we were going to call them in top to bottom order, and noticing that `Math.log(params[0])` appears in both, would in fact generate the code like this:

```scala
def prior(params: Array[Double], globals: Array[Double]): Double = {
   globals(0) = Math.log(params(0))
   globals(0) * 2
}

def likelihood(params: Array[Double], globals: Array[Double]): Double = 
   (globals(0) + params(0)) * params(1)
```

What I've just described above isn't anything to do with this PR, it's just showing how things already work in the compiler.

Now, say that, as I do describe in this PR, we in fact generate two copies of the likelihood function, like this (again showing without any use of globals):

```scala
def prior(params: Array[Double], globals: Array[Double]): Double = 
   Math.log(params(0)) * 2

def likelihoodA(params: Array[Double], globals: Array[Double]): Double = 
   (Math.log(params(0)) + params(0)) * params(1)

def likelihoodB(params: Array[Double], globals: Array[Double]): Double = 
   (Math.log(params(0)) + params(0)) * params(2)
```

The same, existing logic around reusing values would in fact generate something like this:

```scala
def prior(params: Array[Double], globals: Array[Double]): Double = {
   globals(0) = Math.log(params(0))
   globals(0) * 2
}

def likelihoodA(params: Array[Double], globals: Array[Double]): Double = {
   globals(1) = globals(0) + params(0)
   globals(1) * params(1)
}

def likelihoodB(params: Array[Double], globals: Array[Double]): Double =
   globals(1) * params(2)
```

Now, again, this is assuming that we're calling first `prior`, then `likelihoodA`, then `likelihoodB` in sequence, to be semantically correct. But it is equally semantically correct to first call `prior`, then `likelihoodA` (with `params[1]` set to `x[0]`), then `likelihoodB` a large number of times (with `params[2]` set to `x[i]`).

gkossakowski:
This is superb explanation. It makes sense to me now. Very clever use of existing primitives, sir.
I will follow up with the actual review tomorrow.

avibryant:
This currently slows down running the tests by a factor of at least 5x (which is causing it to fail in CI). We shouldn't merge until that's fixed.

avibryant:
Ok, I have pushed a change that will speculatively inline the observation data into the compute graph and bail if the root node is growing too large (since it will ~always be a `Line` node representing a sum of k terms, we can just look at how big k is). In practice this puts all the tests back into the inlined regime and so we're back to the same performance as before. However, it's surprising to me just how much they slowed down previously, and we should do some profiling to see if we can close that gap.

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -1,76 +1,61 @@
...
-        val row: Map[Variable, Real] = data.map {
-          case (v, a) => v -> Real(a(i))
+      var result = Real.zero
+      var i = 0
+      while (i < nRows) {
+        result += inlineRow(i)
+        i += 1
+        result match {
+          case l: Line if (l.ax.size > MAX_INLINE_TERMS) =>
+            println("!" + l.ax.size)
```
gkossakowski: left-over debugging line?

gkossakowski (approved):
lgtm

gkossakowski:
the code looks to me to be doing what you described

a few more questions:
1. what's the motivation for breaking up batched invocation? was the data packing into a matrix and then keeping track of where it went difficult?
2. do you have a sense of what's the typical number of math ops per invocation? i'm wondering where the overhead you found would come from
3. the logic behind this PRs is clever but really tricky. the upper layer (modified in this code) relies on the details of a lower layer (that compiles code). In other words, abstraction layers are violated. I'm wondering if extending IR to do what you want wouldn't be more direct and more maintainable solution. Have you thought about that?

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
avibryant: thanks, fixed

gkossakowski:
one more thought, i can imagine that the answer to the last question is "we don't know the final requirements yet"

avibryant:
@gkossakowski yes, a few answers to the last question:
* we indeed don't know the final requirements yet
* within reason, I believe in keeping lower level primitives simple and general and building on top of them at higher levels; the fact that we *can* do this without changing the lower layer is weak evidence that we should
* I actually started out by trying to do this at the IR level and stopped when I realized it was going to be *significantly* more work and more code and more concepts to do it that way. Which isn't to say that someone else couldn't find a better way to do that, but not me-this-week.

avibryant:
Re your other questions: I am definitely worried that the median number of math ops per invocation is going to be quite low here. The main thing I was really trying to achieve here was to allow us to separate out each group of observations (ie each call to `fit`) into their own compiled function; in particular, I'm doing this while looking forward to hierarchical models (I can explain the connection there but it would take a bit of work to do). However, the fact that in doing so we rolled all the way back to a *single* observation per invocation is not strictly necessary. We can (and, I'm realizing, should) bring back the previous approach of replicating the computation k times so that we are computing batches of k observations at once, which will help amortize the overhead a lot better.

avibryant:
(merging but will follow up to address performance)

Issue #206, opened 2018-09-20
Add `com.stripe.sampler.DensityFunction` trait


avibryant:
As we invest in more complicated compiled representations of the density (allowing batch evaluation etc), it feels more like a leaky abstraction to have the `sampler` package directly interact with the compiler. I propose that we make `sampler` completely dependency-free, and expose a trait like the following, which clients (ie, `core`, making use of `compute`) can supply implementations of:

```scala
trait DensityFunction {
   def arity: Int
   def compute(params: Array[Double]): Double
   def computeGradient(params: Array[Double], gradient: Array[Double]): Double
}
```

All of the `sampler` implementations would be expressed in terms of this, taking it in place of the current `Context` object.

mio-stripe:
Great, I'll start on this.

PR #214, opened 2018-10-02
Decouple Sampler and Compute for batching

mio-stripe:
This PR addresses this issue https://github.com/stripe/rainier/issues/206 by defining a `DensityFunction` trait in Sampler that all samplers use instead of a `Context` so that `Sampler` is dependency-free.


rainier-core/src/main/scala/com/stripe/rainier/compute/Context.scala
```diff
@@ -59,6 +61,14 @@ case class Context(base: Real, batched: Seq[Target]) {
     }
   }
 
+  def compileGradient: Array[Double] => Array[Double] = ???
```
mio-stripe: @avi-stripe this is still `???` Let's chat about it tomorrow.

rainier-core/src/main/scala/com/stripe/rainier/sampler/HMC.scala
```diff
@@ -1,14 +1,13 @@
 package com.stripe.rainier.sampler
 
-import com.stripe.rainier.compute._
 import scala.collection.mutable.ListBuffer
 
 final case class HMC(nSteps: Int) extends Sampler {
-  def sample(context: Context,
+  def sample(densityFun: DensityFunction,
```
avibryant: nit: let's just call this `density`

rainier-core/src/main/scala/com/stripe/rainier/sampler/MAP.scala
```diff
@@ -1,27 +1,27 @@
...
                iterations: Int,
                stepSize: Double): Array[Double] = {
-    val cf =
-      context.compiler
-        .compile(context.variables, context.density :: context.gradient)
-    val initialValues = context.variables.map { v =>
-      0.0
-    }.toArray
+    val cf = (params: Array[Double]) =>
+      densityFunction.density(params) +: densityFunction.gradient(params)
```
avibryant: you're building this list here and then taking it apart again on line 27. could just simplify by calling densityFunction.gradient directly there.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -122,47 +119,69 @@ private[sampler] case class LeapFrog(nVars: Int,
 }
 
 private[sampler] object LeapFrog {
-  def apply(context: Context): LeapFrog = {
-    val qs = context.variables
-    val ps = qs.map { _ =>
-      new Variable
+  def apply(densityFun: DensityFunction): LeapFrog = {
+
+    def split(inputs: Array[Double])
```
avibryant: we need to find a way to do this with zero allocation - previously when we had similar split() logic to this, we were getting hosed with GC when doing enough iterations of large enough modes

avibryant (changes requested):

rainier-core/src/main/scala/com/stripe/rainier/sampler/HMC.scala
mio-stripe: 👍

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: Yeah, I figured we'd have to fix this. Okay, I'll work on this.

rainier-core/src/main/scala/com/stripe/rainier/sampler/MAP.scala
mio-stripe: 👍

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -43,19 +39,17 @@ private[sampler] case class LeapFrog(nVars: Int,
       implicit rng: RNG): Double = {
     initializePs(params)
     params(stepSizeIndex) = stepSize
-    CompiledFunction.run(initialHalfThenFullStep, params, globals, buf2)
+    update(params, buf1)
```
mio-stripe: I'm now copying the input params into `buf1` and doing all the updates to that. I removed the second buffer entirely. Let me know if I'm missing something and this totally overlooks something we were doing before.

mio-stripe:
So I _think_ I'm still doing things correctly in LeapFrog, but I have to implement the `Context.density` to run it.

avibryant:
@mio-stripe it may be worth waiting for my in-progress PR which basically replaces `Context` with batch eval stuff.

rainier-core/src/main/scala/com/stripe/rainier/sampler/MAP.scala
```diff
@@ -1,30 +1,33 @@
...
+        0.0
+      }
+      .toArray
     1.to(iterations).foldLeft(initialValues) {
       case (values, _) =>
-        val _ :: gradient = cf(values).toList
+        val gradient = cf(values).toList
         values.zip(gradient).map {
           case (v, g) =>
             v + (stepSize * g)
```
avibryant: would be pretty easy to replace this with zipWithIndex instead of zip(gradient), and then density.gradient(i), right? looking to get rid of gradientVector if possible

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -122,47 +117,76 @@ private[sampler] case class LeapFrog(nVars: Int,
...
-    val newPotential = context.density * -1
-    val grad = context.gradient.map { g =>
-      g * -1
+  def apply(density: DensityFunction): LeapFrog = {
+
+    val nVars = density.nVars
+    val potentialIndex = nVars * 2
+    val stepSizeIndex = potentialIndex + 1
+
+    def halfPs(inputs: Array[Double]): Unit = {
```
avibryant: could these all just be instance methods on LeapFrog now instead of it having to take them as params? (it would take `density` instead)

rainier-core/src/main/scala/com/stripe/rainier/sampler/MAP.scala
mio-stripe: ah, sure, yeah.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: Yep! Updated.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,23 +1,77 @@
...
   array(n*2+1) == stepSize
    */
+  private val nVars = density.nVars
   private val potentialIndex = nVars * 2
   private val stepSizeIndex = potentialIndex + 1
   private val inputOutputSize = stepSizeIndex + 1
   private val buf1 = new Array[Double](inputOutputSize)
 
+  def halfPs(inputs: Array[Double]): Unit = {
+    density.update(inputs)
```
avibryant: this won't work right because the ps are at the start of the array, right?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,23 +1,77 @@
...
+    var i = 0
+    val j = nVars
+    while (i < j) {
+      inputs(i) = inputs(i) - (inputs(stepSizeIndex) / 2) * density.gradient(i) * -1
```
avibryant: I think `inputs(i) -= ...` is valid syntax here.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,23 +1,77 @@
...
-    twoFullSteps: Array[Double] => Unit,
-    finalHalfStep: Array[Double] => Unit
-) {
+private[sampler] case class LeapFrog(density: DensityFunction) {
   /*
   Params layout:
   array(0..(n-1)) == ps
   array(n..(n*2-1)) == qs
   array(n*2) == potential
   array(n*2+1) == stepSize
```
avibryant: I don't think we need `stepSize` to be part of the input array anymore, right? It can just be passed as a normal param to LeapFrog.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,88 @@
...
   private val inputOutputSize = stepSizeIndex + 1
   private val buf1 = new Array[Double](inputOutputSize)
-  private val buf2 = new Array[Double](inputOutputSize)
 
-  private val globalsSize =
-    List(initialHalfThenFullStep, twoFullSteps, finalHalfStep)
-      .map(_.numGlobals)
-      .max
-  private val globals = new Array[Double](globalsSize)
+  def halfPs(inputs: Array[Double]): Unit = {
```
avibryant: these are always passed `buf1` so now that they're instance methods you don't need them to have a param

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,88 @@
...
+    inputs(potentialIndex) = density.density * -1
+  }
+
+  def finalHalfStep(inputs: Array[Double]): Unit = {
+    halfPs(inputs)
+    density.update(inputs)
+    inputs(potentialIndex) = density.density * -1
+  }
+
+  private def update(sourceArray: Array[Double],
```
avibryant: `copy` might be a clearer name than `update`?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,88 @@
...
+                     targetArray: Array[Double]): Unit =
+    System.arraycopy(sourceArray, 0, targetArray, 0, inputOutputSize)
 
   //Compute the acceptance probability for a single step at this stepSize without
   //re-initializing the ps, or modifying params
   def tryStepping(params: Array[Double], stepSize: Double): Double = {
```
avibryant, line 73: oh right this is where stepsize comes from... so it should be a param to all the `...step...` methods, not an ivar of LeapFrog

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,88 @@
...
   array(0..(n-1)) == ps
   array(n..(n*2-1)) == qs
   array(n*2) == potential
   array(n*2+1) == stepSize
    */
+  private val nVars = density.nVars
   private val potentialIndex = nVars * 2
   private val stepSizeIndex = potentialIndex + 1
   private val inputOutputSize = stepSizeIndex + 1
   private val buf1 = new Array[Double](inputOutputSize)
```
avibryant: nit: now that we only have one of them, call it just `buf`

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: Updated

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: We want to call LeapFrog with different stepSizes though so seems like we could pass it as an argument to the functions that use it rather than looking it up in the array.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: Ah, right, I was imaging the DensityFunction could have the whole big array, but it should definitely only know about the qs.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: 👍

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: yeah, I agree. I'll do that.

mio-stripe:
I added back a `buf2` so we can copy the updated `qs` to it to update the density correctly. I'd still like not to try to put the qs first in the array until I am reasonably confident this is working correctly.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: I think this is good now.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: Ha, `sbt` complained when I had nullary side-effecting methods, but these now take `stepSize` as a param so ✅

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,90 @@
...
   private val potentialIndex = nVars * 2
-  private val stepSizeIndex = potentialIndex + 1
-  private val inputOutputSize = stepSizeIndex + 1
+  private val inputOutputSize = potentialIndex + 1
   private val buf1 = new Array[Double](inputOutputSize)
-  private val buf2 = new Array[Double](inputOutputSize)
+  private val buf2 = new Array[Double](nVars)
+
+  def halfPs(stepSize: Double): Unit = {
+    copyQs()
```
avibryant: It wasn't clear to me that `copyQs` would also update `density`.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,90 @@
...
+    var i = 0
+    val j = nVars
+    while (i < j) {
+      buf1(i) -= (stepSize / 2) * density.gradient(i) * -1
```
avibryant: nit: this could be `+=` and drop the `* -1`, right?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,90 @@
...
+  private val nVars = density.nVars
   private val potentialIndex = nVars * 2
-  private val stepSizeIndex = potentialIndex + 1
-  private val inputOutputSize = stepSizeIndex + 1
+  private val inputOutputSize = potentialIndex + 1
   private val buf1 = new Array[Double](inputOutputSize)
-  private val buf2 = new Array[Double](inputOutputSize)
+  private val buf2 = new Array[Double](nVars)
+
+  def halfPs(stepSize: Double): Unit = {
```
avibryant: hey, we don't need both `halfPs` and `fullPs`, we can just call `fullPs` with `stepSize / 2.0`

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,90 @@
...
+    val j = nVars
+    while (i < j) {
+      buf1(i) -= (stepSize / 2) * density.gradient(i) * -1
+      i += i
+    }
+  }
+
+  def halfPsNewQs(stepSize: Double): Unit = {
+    halfPs(stepSize)
+    var i = nVars
```
avibryant, line 17: we can pull the `newQs` part out since we use it here and also in `fullPsNewQs`? And in fact then we probably don't need the methods like `halfPsNewQs` we can just call `newPs(stepSize / 2.0); newQs(stepSize)`.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -71,8 +120,7 @@ private[sampler] case class LeapFrog(nVars: Int,
 
   //we want the invariant that a params array always has the potential which
   //matches the qs. That means when we initialize a new one
-  //we need to compute the potential. We can do that (slightly wastefully)
-  //by using initialHalfThenFullStep with a stepSize of 0.0
+  //we need to compute the potential.
```
avibryant, line 112: nice catch updating this comment

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -1,38 +1,90 @@
...
-  array(n*2+1) == stepSize
    */
+  private val nVars = density.nVars
   private val potentialIndex = nVars * 2
-  private val stepSizeIndex = potentialIndex + 1
-  private val inputOutputSize = stepSizeIndex + 1
+  private val inputOutputSize = potentialIndex + 1
   private val buf1 = new Array[Double](inputOutputSize)
-  private val buf2 = new Array[Double](inputOutputSize)
+  private val buf2 = new Array[Double](nVars)
```
avibryant: maybe instead of buf1 and buf2 we can call them `pqBuf` and `qBuf`?

avibryant:
minor tweaks aside, this is looking pretty good. I think maybe we should put it on pause, get in the rest of the batch eval work tomorrow (I can finish it on the plane), then rebase this onto that and merge from there

mio-stripe:
Sounds good: I'll make these tweaks and then rebase onto your eval stuff.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
mio-stripe: Fixed

avibryant:
Ok cool. This looks ready to me for combining with https://github.com/stripe/rainier/pull/215 . I can probably do that tomorrow if you want, or if you want to do it let me know.

mio-stripe:
I rebased and changed the base branch to `targetvector2` so this should be good to go.

avibryant:
@mio-stripe I've hooked up `DensityFunction` to my `DataFunction` but so far we're getting broken results (HMC hangs and Walkers outputs crap). To be debugged...

avibryant (approved):

avibryant:
Found the bug, waiting on CI and then kicking this back to you, @mio-stripe

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -117,7 +109,23 @@ class RandomVariable[+T](val value: T, private val targets: Set[Target]) {
     (allSamples, diagnostics)
   }
 
-  lazy val density: Real = Real.sum(targets.map(_.inlined))
+  lazy val (variables, density) = compile()
+  private def compile() = {
+    val (variables, dataFn) = Compiler.default.compileTargets(targets, true, 4)
```
avibryant: Somewhat wasteful that we always compute gradients even when using `Walkers` but I'm not going to worry about that for now.

avibryant:
Oh yeah, build failure on rainierBenchmark, willfix.

avibryant:
This is now failing only on the `tut` build. I'm tempted to just fudge that somehow and come back to it as part of a more major docs revision.

mio-stripe:
Cool! That sounds reasonable. Can we comment breaking tut stuff out and merge into develop since the out-of-date-but-not-test-failing docs are still on master?

mio-stripe:
Nice, 🔪 `Context`.
LGTM

PR #215, opened 2018-10-04
DataFunction

avibryant:
This introduces the `DataFunction` abstraction which provides a standard representation of computations that operate on one or more matrices of data. It wraps up a very carefully and specifically constructed `CompiledFunction` along with a sequence of 2D data matrices and takes care of the efficient evaluation of one or more outputs that involve mapping a calculation over the matrices and summing the results.

This also adds a `compileTargets` method to `Compiler` which takes a set of `Target` objects and prepares a `DataFunction` from them; `Context` has been adapted to use this (currently just for the simple non-gradient case).

There are two new things we're doing in `DataFunction` relative to the approach in https://github.com/stripe/rainier/pull/213:

* first, we are allowing multiple outputs rather than just one, in anticipation of getting gradients out in the future. This is a straightforward extension.
* second, we are allowing multiple observations to be computed and summed within a single invocation of a compiled function, to cut down on overhead. Specifically: `compileTargets` and `DataFunction` both take a `batchBits` argument, and rather than producing a *single* copy of each likelihood function, we produce a copy for each of `k in (0,batchBits)`, where each copy processes `2^k` observations at once. Thus we can process data in batches of size 2^batchBits at a time, until we are left with < 2^batchBits items, at which point we can make at most batchBits calls to the smaller-batch copies to process the remainder.


rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
```diff
@@ -20,6 +20,58 @@ trait Compiler {
...
+                     batched: List[Target],
+                     gradient: Boolean,
+                     batchBits: Int): DataFunction = {
+
+    val variables =
+      batched
+        .foldLeft(RealOps.variables(base)) {
+          case (set, target) =>
+            set ++ target.variables
+        }
```
gkossakowski, line 40: isn't this `RealOps.variables(base).toSet ++ batched.map(_.variables)` ?

gkossakowski: Oh I see below you want a List, then `(RealOps.variables(base) ++ batched.map(_.variables)).distinct` would be more direct than a foldLeft.

rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
```diff
@@ -20,6 +20,58 @@ trait Compiler {
...
+        .sortBy(_.param.sym.id)
+
+    def withGradient(real: Real): List[Real] =
+      if (gradient)
+        real :: Gradient.derive(variables, real)
+      else
+        List(real)
+
+    val (batchVariables, batchOutputs) =
+      batched.foldLeft((List.empty[Variable], List.empty[Real])) {
```
gkossakowski, line 61: I would pick `flatMap(...).unzip` over the foldLeft but it's a weak preference

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
```diff
@@ -47,6 +47,29 @@ class Target(val real: Real, val placeholders: Map[Variable, Array[Double]]) {
...
+  }
+
+  private def batch(bit: Int): (List[Variable], Real) =
+    0.until(1 << bit).foldLeft((List.empty[Variable], Real.zero)) {
+      case ((v, r), _) =>
+        val newVars = placeholderVariables.map { _ =>
+          new Variable
+        }
+        val newReal =
+          PartialEvaluator.inline(real, placeholderVariables.zip(newVars).toMap)
```
gkossakowski, line 69: In the PR comment you mention that each copy of the function is restricted to computing a 2^k observations but I don't see how this is implemented. Do I see it right that a single batch creates `2^bit` copies of exactly the same function?

If so:

1. Wouldn't it be more direct to say?
```
List.fill(2^bit) {
  val newVars = ...
  ...
}
```

2. is there any concern over code size blowup? or is packing of IR supposed to take care of it by reusing most of the IR?

rainier-core/src/main/scala/com/stripe/rainier/ir/DataFunction.scala
```diff
@@ -0,0 +1,156 @@
+package com.stripe.rainier.ir
+
+import scala.annotation.tailrec
+
+/*
+Input layout:
+- numParamInputs param inputs
+- for 0 <= i < data.size:
+   - data[i].size first-run inputs
```
gkossakowski, line 9: what does "first-run inputs" mean?

rainier-core/src/main/scala/com/stripe/rainier/ir/DataFunction.scala
```diff
@@ -0,0 +1,156 @@
...
+  def apply(inputs: Array[Double],
+            globals: Array[Double],
+            outputs: Array[Double]): Unit = {
+    computeWithoutData(inputs, globals, outputs)
+    var i = 0
+    while (i < data.size) {
+      computeFirstWithData(inputs, globals, outputs, i)
+      computeRestWithData(inputs, globals, outputs, i)
+      i += 1
+    }
```
gkossakowski, line 52: do I see it right that `computeWithoutData` is the `prior` we discussed in the previous PR?
and:
 `computeFirstWithData` - `likelihoodA`
 `computeRestWithData` - `likelihoodB`
?

gkossakowski (commented):
i _think_ i understand the changes and they mostly look good to me except for some low-level collection operations nitpicking. personally, I like to use the least general collection operations that can do the job and `foldLeft` is the most general one

other than that, I'm concerned with `data: Array[Array[Array[Double]]]` and all of index fiddling going on. i wonder if wrapping the naked nested array in some "view-like" abstraction would clarify what's going on. I have this feeling you have a 3-d picture somewhere in your paper notes that visualizes the memory layout and the current code structure doesn't capture it

gkossakowski:
forgot to say: the idea of generating copies of the some code with exponentially increasing input size as a solution to batching is great

rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
avibryant: I think you're missing a `flatten`, right? but yes, will make the change

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
avibryant: Yes, it's 2^bit copies whose results are then summed.
1. yep, that would be cleaner, thanks
2. I am indeed a bit worried about code size blowup. Depending on the exact function it may be able to optimize or reuse large parts of it, but there's no guarantee of that. In my quick empirical testing, you get most of the benefit from just 2^4 copies, which doesn't seem *that* bad.

rainier-core/src/main/scala/com/stripe/rainier/ir/DataFunction.scala
avibryant: Correct.

rainier-core/src/main/scala/com/stripe/rainier/ir/DataFunction.scala
avibryant: `likelihoodA`

avibryant:
Thanks. I'll see what I can do about your concerns about the naked nested array.

rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
gkossakowski: Yes, or `flatMap`

rainier-core/src/main/scala/com/stripe/rainier/compute/Target.scala
gkossakowski: If a small number of copies is enough, does it make sense to cap them?

mio-stripe:
LGTM

PR #216, opened 2018-10-07
cover all IR cases in packer

avibryant:
We were using a default case when traversing the IR tree in Packer, which was obscuring the fact that we weren't handling `IfIR`, leading methods to be too big if there were large sub-trees under an if. This fixes that and explicitly enumerates all of the cases so we'll notice if we ever add another node type.


mio-stripe:
Nice, LGTM

gkossakowski:
lgtm too (forgot about this pr)

PR #217, opened 2018-10-08
introduce some stan benchmarks

avibryant:
Starting with three of the benchmarks from https://github.com/stan-dev/stat_comp_benchmarks . On two of them we do much better, on the third one we do worse.


rainier-example/src/main/scala/com/stripe/rainier/example/stan/LowDimGaussMix.scala
```diff
@@ -0,0 +1,304 @@
...
+      val mean = params.map { _(k) }.sum / params.size
+      println(s"$k = $mean")
+    }
+  }
+
+  val warmups = 1000
+  val samples = 10000
+  val thin = 10
+
+  val ys = List(-3.58542974746427, 3.11625192303286, 3.55429942030185,
```
gkossakowski: as a rule of thumb it's good to provide a type argument `Double` explicitly when a large number of elements are involved. It helps to drive type inference.

mio-stripe:
Cool! LGTM

PR #224, opened 2018-10-12
introduce Coefficients

avibryant:
This abstracts out the `ax` vectors from `Line` and `LogLine` into a `Coefficients` trait, to better encapsulate that internal representation and allow us to have more sophisticated handing.

The `Coefficients` trait currently has three concrete implementations:
- `Empty` which is only used as an intermediate value (it's illegal to actually build a `Line` with it)
- `One` for the extremely common case of a single-term `Line` or `LogLine`
- `Many` for all other cases. `Many` is implemented as a combination of a `Map` for efficient lookups, along with a separate `List` of the keys for deterministic ordering. Some care was taken to avoid quadratic behavior in almost all cases - the main exception being the `withComplements` method used when building gradients of `LogLine`, but there are already reasons that `LogLine` must and will stay small.


rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
```diff
@@ -0,0 +1,132 @@
...
+        Coefficients(pair :: toList)
+      }
+  }
+
+  case class Many(toMap: Map[NonConstant, BigDecimal], terms: List[NonConstant])
+      extends Coefficients {
+    val isEmpty = false
+    def size = toMap.size
+    def coefficients = toMap.values
+    def toList = terms.map { x =>
```
mio-stripe, line 73: Is this more efficient than `toMap.toList`?

rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
```diff
@@ -0,0 +1,132 @@
...
+      x -> toMap(x)
+    }
+
+    def mapCoefficients(fn: BigDecimal => BigDecimal) =
+      Many(toMap.map { case (x, a) => x -> fn(a) }, terms)
```
mio-stripe, line 78: Is there a reason to prefer explicitly writing `x -> fn(a)` over doing `Many(toMap.mapValues(fn), terms)`?

rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    def withComplements = {
```
mio-stripe, line 80: Cool, this took a little time, but I get it now: we are adding `b.head` to `a` so that that term is included in the complements of all of the other terms. Nice.

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
```diff
@@ -38,18 +38,19 @@ private object Gradient {
...
 
           case l: Line =>
-            l.ax.foreach {
+            val ax = l.ax.toList
+            ax.foreach {
               case (x, a) =>
                 diff(x).register(ProductDiff(a, diff(l)))
             }
-            l.ax.foreach { case (x, _) => visit(x) }
+            ax.foreach { case (x, _) => visit(x) }
```
mio-stripe: Looks like this could be `ax.foreach{ case (x,a) => diff...; visit(x) }`.

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
```diff
@@ -38,18 +38,19 @@ private object Gradient {
...
           case l: LogLine =>
-            l.ax.foreach {
-              case (x, _) =>
-                diff(x).register(LogLineDiff(l, diff(l), x))
+            l.ax.withComplements.foreach {
+              case (x, a, c) =>
+                diff(x).register(LogLineDiff(diff(l), x, a, c))
             }
-            l.ax.foreach { case (x, _) => visit(x) }
+            l.ax.toList.foreach { case (x, _) => visit(x) }
```
mio-stripe: Same comment about combining the `foreach`s

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
mio-stripe: Oh, probably we want to register all of the diffs before we call visit on each term and that's why the two `foreach` are separate.

mio-stripe (commented):
Looks great. Left a few comments/questions.

rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
avibryant: It's likely less efficient but, critically, it's deterministic in the order of its output. (Actually a TODO is to audit the uses of this toList and change them to toMap where they don't actually care about order).

rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
avibryant: Fun Scala fact: `mapValues` creates a wrapper which will lazily apply `fn` on every use of the map, whereas `map` is eager. I'm always scared of the performance implications of storing the result of`mapValues`, especially where multiple of them can stack up.

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
avibryant: Yeah, you're correct about my original intent, though as I think about it again I don't think it actually matters and I probably could in fact combine them. I'll think about it more.

rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
mio-stripe: Riiight, yes. Good todo, also.

rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
mio-stripe: Whoa, TIL 🌠

mio-stripe:
👍

PR #225, opened 2018-10-15
remove the adhoc tests in factor of sbc tests

avibryant:
Now that we have `SBCTest`, we can get rid of these more adhoc versions. We may be losing coverage in the short term but we should build that back up on the SBC side.


mio-stripe:
🎉

PR #232, opened 2018-10-16
conveniences for Regression example

avibryant:
A few things I've been meaning to add that are unrelated except that each of them makes the `Regression` example a bit nicer:

* `RandomVariable.fill` which is a `RandomVariable.traverse` combined with a `List.fill`
* `Real.dot` for dot products (not adding a new node type, just the convenience method)
* a `Mapping` from `Seq[Double]` to `Seq[Real]` (or any other Seq[T], Seq[U] where we have a mapping between T and U)
* a `Predictor.fromDoubleVector` which makes use of the above


mio-stripe:
Nice

PR #237, opened 2018-10-18
smarter packer

avibryant:
The packer is responsible for splitting up the method tree so that no individual subtree exceeds the maximum size we can JIT.

Before, for an n-ary operation (like If, which is ternary), we would evenly divide the method size allocation between each branch. This could lead to unnecessarily creating a new method when the branches were different sizes. Now, if the first branch is small, that will increase the allocation we give to the subsequent branches (and vice versa if the first branch is extremely large).


gkossakowski (approved):

gkossakowski:
i mixed my two GH accounts all the time, i tend to be logged in with my gkossakowski account, though

PR #239, opened 2018-10-19
allow DensityPlot to output raw histogram data for other plotting tools

kailuowang:
This is a simple refactoring that exposes the intermediate data needed for console plotting so that we can easily plot with other tools (like on web).


avibryant:
Thanks!

PR #240, opened 2018-10-19
set the version to SNAPSHOT so that publishing snapshot would be easy.

kailuowang:


kailuowang:
This build error is interesting. rainier-plot is complaining it couldn't find rainierCore0.1.4-SNAPSHOT from maven. This makes me suspect that rainier-plot has been tested against rainier-core 0.1.0 all along, which is probably not good. 
We encountered a similar issue at cats-collections too. It's due to the fact that rainier-plot has a different `crossScalaVersions` than rainier-core, and somehow sbt decides to resolve to remote repo for the core dependency. 
I will ask in the sbt gitter chat room

avibryant:
I get a bit lost in the sbt stuff but could this also be related? https://github.com/stripe/rainier/pull/196 cc @andyscott who's good at this.

kailuowang:
ha, looks like the exact same error with the same version change.  
I asked the question in gitter
https://gitter.im/sbt/sbt?at=5bca1bd0aed22a514b61e285
Also created a minimized example https://github.com/kailuowang/intermodule

kailuowang:
closing per https://github.com/stripe/rainier/pull/241#issuecomment-432072033

PR #241, opened 2018-10-19
updated scala version

kailuowang:


build.sbt
```diff
@@ -9,6 +9,7 @@ lazy val root = project.
   settings(commonSettings).
   settings(unpublished).
   settings(
+    crossScalaVersions := Nil,
```
avibryant, line 12: Interesting, what's the effect of this?

kailuowang: According to Eugene Yokota:
>the subproject should declare whatever the Scala versions it supports. if you have anything on root it will call `compile` on all the subproject that it aggregates, and results to the error you saw.

avibryant: Cool, thanks. This LGTM but let's wait for the github webhook incident to pass so we have CI verification on this before merging.

kailuowang:
Please hold. I realized I need to verify release works with this setting

kailuowang:
I'd like to consolidate this and #240, closing both and will submit a new consolidated PR later.

PR #243, opened 2018-10-19
Better types for IR

avibryant:
This is a large diff because some things get renamed, but a relatively simple idea. The `IR` trait was not properly capturing as much structure as actually existed in the expression trees, and this PR tightens up the types a bit to fix that. Specifically:

* The `IR` trait has been narrowed in focus to only cover the nodes that operate on other values: currently `BinaryIR`, `UnaryIR`, and `IfIR`. There's probably a better system of names for this  but keeping it reduced the size of this diff.
* The main top-level trait is now called `Expr` instead of `IR`. In general, things should now be dealing with `Expr` values rather than `IR` which are now more of an internal implementation detail.
* Under `Expr` is `Ref`, which still has the same subtypes as before.
* The only other `Expr` is `VarDef`. It holds onto an `IR` on its right-hand side. (This illuminates what it means for something to be `IR`: "an operation expensive enough to be worth caching in a var").
* To close the loop, the recursive members of the IR nodes are `Expr` values.

We also pulled `MethodDef` out of this hierarchy entirely since it's really a separate thing, and needed to add in a `NoOp` unary op to paper over an edge case.

The actual logic everywhere should be unchanged, it's just been adapted to the new types.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Compiler.scala
@@ -95,10 +95,10 @@ final case class IRCompiler(methodSizeLimit: Int, classSizeLimit: Int)
     val params = inputs.map { v =>
       v.param
     }
-    val irs = outputs.map { r =>
-      translator.toIR(r)
+    val exprs = outputs.map { r =>
+      translator.toExpr(r)
     }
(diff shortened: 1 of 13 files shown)


avibryant:
cc @gkossakowski in case you want to take a look

mio-stripe:
LGTM

PR #244, opened 2018-10-20
add SeqIR node

avibryant:
This adds the `SeqIR` node which has two children, `first` and `second`, which it runs in sequence and then returns the value of `second`. This seems like a strange thing to have in a pure expression language - what's the point of running `first` if we're throwing away the result? The goal here is effectively to act as a "cache warmer" - in particular, for (future) large n-ary nodes like `Lookup` and `Sum`, it's useful to schedule the computation of each of their children in advance so that the code they ultimately generate can be as compact as possible (just referencing pre-computed values), which means we can pack as large a node as possible into a single JIT-friendly method.

Because of this, `first` and `second` are constrained to be `VarDef`, since there's no point in pre-computing a `Ref`.


avibryant:
Added a WIP because as I'm trying to use this I'm becoming unsure about the types.

rainier-core/src/main/scala/com/stripe/rainier/ir/Tracer.scala
```diff
@@ -97,6 +97,7 @@ object Tracer {
             s"if($t == 0.0) $z else $nz"
           else
             s"(if($t == 0.0) $z else $nz)"
+        case s: SeqIR => ???
```
mio-stripe: This goes away with the new tracer, right?

avibryant: correct, which is why I left it this way :)

avibryant: I'll do the rebase.

avibryant:
Ah actually merging the Lookup branch in fixed the need for a rebase :)

PR #245, opened 2018-10-22
Replace the old Tracer with one based on the CFR decompiler

avibryant:
This introduces a new Tracer that uses the https://www.benf.org/other/cfr/ decompiler to produce Java code from the actual generated bytecode rather than trying to simulate this from the IR.

Since it introduces a new dependency, and is mostly useful during development, it's in a separate module that is not published. Since CFR is only distributed as a jar, and isn't on maven, it's included as a binary jar in this PR.

In order to integrate it, I had to decompile it (with itself!) to read enough of the source to understand the API. That also gives me some comfort about including it even though there's no published source, though it raises the amusing spectre of "Reflections on trusting trust" (but backwards!). (In practice the risks here are no different than any other binary dep).


mio-stripe:
This is very cool.
LGTM

PR #247, opened 2018-10-22
Lookup

avibryant:
Depends on both #244 and #245 (which means that #245's changes appear in this diff for now).

This adds a new `Real` type, `Lookup(index: Real, table: Seq[Real])`, which computes `table(index)`. The expected (only?) use case is within a `Predictor` where you have an integer component of your observations that you want to use to select from a vector of parameters. For example: picking the parameter specific to a sub-population in a hierarchical model, or picking the parameter specific to a time-period in a time-series model.

Currently this translates to IR as follows:
* for anything in `table` whose IR is a `VarDef` (ie this is the first time we've computed this value), generate a `SeqIR` tree that warms the cache for it (packed into multiple generated methods as necessary)
* in the final result location of the SeqIR tree a single `LookupIR` which holds `Ref`s to everything in `table`, which generates a single `switch` statement for the entire table; if packed alone into a method, it would be roughly 2-3x as many bytecodes as `table.size`. There is no guarantee that this will be small enough to JIT, but in almost all cases it will be, and if we're seeing slowdowns from it not being JITted we can work on splitting it up.


mio-stripe:
LGTM

PR #250, opened 2018-10-23
fix build with updated deps

kailuowang:
* updated scala and scalafix
* updated evilPlot which supports scala 2.11 now
* removed 2.12 only crossScalaVersions


kailuowang:
this should fix #196 as well.

avibryant:
Can you merge or rebase this against develop? That should make CI pass.

avibryant:
Thanks for all your work on this!

PR #252, opened 2018-10-23
add dep on google flogger

avibryant:
As a first step towards having better logging, this adds a dep on google's flogger and demonstrates using it by creating an instance for the `compute` package and adding a log message in `Compiler.compile`.

I don't yet actually know how you set the log level but presume there's some standard java way of doing this. I'm also violating the expectations a bit by allocating the logger in the package object vs in the companion object of each class, which would have given more finegrained control but was too much boilerplate for me (especially in Scala, it's better in Java).


mio-stripe:
Flogger: the self-flagellating logger :)
LGTM

PR #253, opened 2018-10-24
Benchmarks with MAX_INLINE=500 and 100 observations

mio-stripe:


avibryant:
LGTM. Note: GeometricZeroInflated and Laplace at 1000 get slower, which is expected (they're transitioning from a single compute graph to batched eval).

PR #254, opened 2018-10-26
Visualize Real graphs with GraphViz

avibryant:
This adds `writeGraph`and `writeIRGraph` methods to both `Real` and `RandomVariable` which output files in GraphViz format which can be visualized with the `dot` tool. Here's an example of the output for the following RV:

```scala
val data = 1.to(1000).map{i => i.toDouble}.toList
Normal(0, 1).param.flatMap{x =>
   Laplace(x,1).fit(data)
}
```

The Real graph:
real

The IR graph:
ir


rainier-core/src/main/scala/com/stripe/rainier/ir/GraphViz.scala
```diff
@@ -0,0 +1,77 @@
+package com.stripe.rainier.ir
```
avibryant: this is in the IR package because I intend to add an `ir.Viz` as well that makes us of it to visualize IR trees

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -136,6 +136,24 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
   def toGenerator[U](
       implicit tg: ToGenerator[T, U]): RandomVariable[Generator[U]] =
     new RandomVariable(tg(value), targets)
+
+  def writeGraph(path: String, maybeInline: Boolean = false): Unit = {
```
avibryant: might be clearer to call this `writeDensityGraph`

avibryant:
This is what it would look like with all the `x` stuff gone.
foo gv

rainier-core/src/main/scala/com/stripe/rainier/compute/Viz.scala
```diff
@@ -0,0 +1,123 @@
...
+        arr.take(5).toList.map(double)
+    }
+    val colIDs = gv.matrix(mid, "data", colData)
+    cols.zip(colIDs).foreach {
+      case ((v, _), cid) =>
+        ids += (v -> cid)
+    }
+  }
+
+  def double(c: Double): String =
```
avibryant: we should be smarter about what we do here

mio-stripe:
Cool, graphs and code look good. It's not obvious what the symbols in the graphs mean without looking in the code so we'll need to add (later) some documentation to make the graphs comprehensible.

PR #255, opened 2018-10-30
ir.Viz

avibryant:
This adds `writeIRGraph` to `Real` so that we can also visualize IR trees. Currently there's no equivalent for RandomVariable but we can add that as a follow up. Here's a simple example:

foo gv


avibryant:
Merging this into #254 since I'm finding myself iterating on both in tandem.

PR #256, opened 2018-10-30
Fold over sum

mio-stripe:
This is an alternative to adding a SumIR node: we do a foldLeft over a sum of terms creating binary nodes which we memoize as we go.


rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
```diff
@@ -148,9 +171,31 @@ private class Translator {
...
 
+  private def combineTerms(terms: Seq[(Real, BigDecimal)], ring: Ring): Expr = {
+    val lazyExprs = makeLazyExprs(terms, ring)
+    combineTree(lazyExprs, ring)
+  }
+
+  private def combineSumTerms(terms: Seq[(Real, BigDecimal)]): Expr = {
+    val ring = multiplyRing
+    val lazyExprs = makeLazyExprs(terms, ring)
+    foldChain(lazyExprs, Const(0.0), 0)
```
mio-stripe: We don't actually care about the initial value of the accumulator: we keep track of which iteration we are on and replace it with the first term in the list on iteration 0.

mio-stripe:
even with N = 50 observations, the generated code looks totally reasonable, but at N = 1000 I get an overflow

```
Oct 30, 2018 10:47:37 AM com.stripe.rainier.compute.Compiler compile
INFO: Compiling method with 1 inputs and 1 outputs
java.lang.StackOverflowError
  at com.stripe.rainier.ir.Packer.traverseIR(Packer.scala:32)
  at com.stripe.rainier.ir.Packer.traverseVarDef(Packer.scala:19)
  at com.stripe.rainier.ir.Packer.traverse(Packer.scala:14)
  at com.stripe.rainier.ir.Packer.traverseIR(Packer.scala:32)
  at com.stripe.rainier.ir.Packer.traverseVarDef(Packer.scala:19)
  at com.stripe.rainier.ir.Packer.traverse(Packer.scala:14)
  at com.stripe.rainier.ir.Packer.traverseIR(Packer.scala:32)
  at com.stripe.rainier.ir.Packer.traverseVarDef(Packer.scala:19)
  at com.stripe.rainier.ir.Packer.traverse(Packer.scala:14)
  at com.stripe.rainier.ir.Packer.traverseIR(Packer.scala:32)
  at com.stripe.rainier.ir.Packer.traverseVarDef(Packer.scala:19)
  at com.stripe.rainier.ir.Packer.traverse(Packer.scala:14)
```

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
```diff
@@ -96,6 +96,28 @@ private class Translator {
   The result may also be multiplied by a constant scaling factor (generally
   factored out of the original summation).
   **/
+  private def factoredSumLine(ax: Coefficients,
+                              b: BigDecimal,
+                              factor: Double): Expr = {
+    val ring = multiplyRing
```
avibryant: is there any point to using `ring` here instead of just inlining all of the ops/values below? It seems like unnecessary indirection.

rainier-trace/src/main/scala/com/stripe/rainier/trace/Tracer.scala
```diff
@@ -21,6 +21,13 @@ case class Tracer(compiler: Compiler, gradient: Boolean) {
       compiler.compileTargets(rv.targets, gradient, batchBits)
     Tracer.dump(df.cf)
   }
+
+  def output(rv: RandomVariable[_], batchBits: Int = 1): String = {
```
avibryant: so this returns a string rather than printing it - out of curiosity, where did you find that useful? Maybe we can find a better abstraction that the two-level code duplication here.

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
avibryant: Weird :) Is it necessary? And why can't we just use a normal foldLeft?

mio-stripe: Normal foldLeft works, but it adds an extra `0 + ` to every sum because it doesn't throw out the accumulator on the first iteration. We could just do that.

rainier-trace/src/main/scala/com/stripe/rainier/trace/Tracer.scala
mio-stripe: When I was getting huge trace output last week, I think, it was too much to display in the repl so  I added this so I could capture all of it. Yeah, I'll think about a neater way to do this.

rainier-benchmark/benchmarks.txt
```diff
@@ -112,8 +112,8 @@ run                  BinomialPoissonApproximation 10000
 run                  BinomialPoissonApproximation 100000                         1.000
 run                  Exponential               100                            3.000
 run                  Exponential               1000                          50.000
-run                  Exponential               10000                        400.000
-run                  Exponential               100000                      4000.000
+run                  Exponential               10000                        500.000
+run                  Exponential               100000                      5000.000
```
mio-stripe, line 116: This is not great

rainier-benchmark/benchmarks.txt
```diff
@@ -122,9 +122,9 @@ run                  Geometric                 100                            1.
 run                  Geometric                 1000                           1.000
 run                  Geometric                 10000                          1.000
 run                  Geometric                 100000                         1.000
-run                  GeometricZeroInflated     100                            1.000
-run                  GeometricZeroInflated     1000                        1000.000
-run                  GeometricZeroInflated     10000                      10000.000
+run                  GeometricZeroInflated     100                            2.000
```
mio-stripe, line 125: Down from 20x to 2x!

rainier-benchmark/benchmarks.txt
```diff
@@ -122,9 +122,9 @@ run                  Geometric                 100                            1.
...
+run                  GeometricZeroInflated     1000                         300.000
+run                  GeometricZeroInflated     10000                       3000.000
```
mio-stripe, line 127: ...and these are faster??

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
mio-stripe: Inlining is better. Fixed.

rainier-benchmark/benchmarks.txt
avibryant: overall I think this all reads to me as "no change" given the still-imperfect stability of the benchmark numbers. which is good.

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
avibryant: Could this be 

```scala
lazyExprs.tail.foldLeft(lazyExprs.head){
(acc, t) => binaryExpr(acc, t(), AddOp)
}
```

avibryant:
My bet is that N=1000 works now and caused a stack overflow before is that in the meantime you've merged something in that switches the threshold from inlined data -> batch data from its previous value of 2000, to its new value of 500?

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
```diff
@@ -96,6 +96,27 @@ private class Translator {
...
+      case -1.0 =>
+        binaryExpr(Const(0.0), expr, SubtractOp)
+      case 2.0 =>
+        binaryExpr(expr, ref(expr), AddOp)
+      case k =>
+        binaryExpr(expr, Const(k), MultiplyOp)
+    }
+  }
+
   private def factoredLine(ax: Coefficients,
```
avibryant, line 120: I feel like there's a bunch of simplification to do here, since we now have separate code paths for times and plus - eg `factoredLine` will never take a non-identity `b` term now. Probably not for this PR but should keep as a TODO.

rainier-trace/src/main/scala/com/stripe/rainier/trace/Tracer.scala
avibryant: Maybe just drop it from this PR for now.

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
mio-stripe: yes, definitely :)

rainier-benchmark/benchmarks.txt
mio-stripe: Yeah, that makes sense.

mio-stripe:
Ugh, I think you're right: I merged https://github.com/stripe/rainier/pull/253 to get the updated benchmarks and didn't think about that it obviously has the MAX_INLINE change.

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
mio-stripe: Ah, very good point.

mio-stripe:
I'm going to update the SBC's.

avibryant:
If you're happy with this now let's merge it and we can do further cleanup in Translator as you do the kahan stuff.

mio-stripe:
Sounds like a plan.

PR #257, opened 2018-10-31
Use median for benchmarks

avibryant:
This uses an aggressively rounded median for benchmarks instead of mean, which is considerably more stable and should lead to fewer spurious diffs.


mio-stripe:
LGTM

PR #259, opened 2018-11-02
Fix lookup

avibryant:
I'm really confused as to how this ever worked, but: Lookup has been broken because it tries to do a `tableswitch` with a `double` value instead of an `int`. This adds the conversion instruction so that things work again.


mio-stripe:
LGTM

PR #260, opened 2018-11-02
get rid of If in favor of Compare+Lookup

avibryant:
The new `Compare` node more directly models the JVM instruction set, and when combined with Lookup gives us what should be a more efficient implementation of branching on comparisons than we had previously, and with fewer primitives. Still needs a bit of cleanup and validation via benchmarks that it is in fact faster.


rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -99,9 +99,7 @@ object Multinomial {
     Combinatorics.factorial(multi.k) + Real.sum(v.toList.map {
       case (t, i) =>
         val p = multi.pmf.getOrElse(t, Real.zero)
-        val pTerm =
-          If(i, i * p.log, If(p, whenNonZero = i * p.log, whenZero = Real.zero))
-
+        val pTerm = i * p.log //???
```
mio-stripe: Looking at this again, it took me a while to refigure out that `pmf` is the map of outcomes to their probabilities and `v` is the map of outcomes to their number of occurrences. An extra sentence in the comment on the role of `value` in the `logDensity` would help.

rainier-benchmark/benchmarks.txt
```diff
@@ -122,11 +121,11 @@ run                  Geometric                 100                            1.
 run                  Geometric                 1000                           1.000
 run                  Geometric                 10000                          1.000
 run                  Geometric                 100000                         1.000
-run                  GeometricZeroInflated     100                            1.000
```
avibryant, line 125: why'd this get slower?

avibryant:
I think this is worth merging at this point, though could certainly use more documentation and optimization.

mio-stripe:
Whoops, forgot to check back on this. Looking now.

mio-stripe:
Cool, LGTM

mio-stripe:
Looks like just the expected SBCs to update

PR #261, opened 2018-11-09
Make logging useful

avibryant:
This adds:

* a `log.Logger` trait which has methods for `INFO`, `FINE`, etc that return flogger `Api` objects for logging with; it also includes `showInfo`, `showFine`, etc for programmatic control of what gets shown on the console (logging to files etc can be controlled by config files)
* `sampler.Log`, `core.Log`, `compute.Log` etc instances of `Logger` to give per-package control over logging
* the convention of `import Log._` in classes that want to produce log statements so that they can just say `FINE.log(....)`
* a smattering of examples of using the above

Note that we have to do some annoying tricks with reflection to get around the mismatch between the assumptions flogger has about being used from java and the realities of using it from scala.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/sampler/Walkers.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/sampler/Walkers.scala
@@ -2,29 +2,41 @@ package com.stripe.rainier.sampler
 
 import scala.collection.mutable.ListBuffer
 
+import Log._
+import java.util.concurrent.TimeUnit._
+
 final case class Walkers(walkers: Int) extends Sampler {
   def sample(density: DensityFunction,
              warmupIterations: Int,
              iterations: Int,
              keepEvery: Int)(implicit rng: RNG): List[Array[Double]] = {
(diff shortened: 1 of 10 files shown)


avibryant:
It's worth noting that the logging settings persist during an sbt session, and so you can drop into `console` to `showFine()` and then `run` to see the log statements show up.

mio-stripe:
Nice

rainier-core/src/main/scala/com/stripe/rainier/sampler/Walkers.scala
```diff
@@ -2,29 +2,41 @@ package com.stripe.rainier.sampler
...
             chain.next
         }
 
     val buf = new ListBuffer[Array[Double]]
     var i = 0
     var chain = warmedUp
+
+    FINE.log("Starting %d iterations", iterations)
     while (i < iterations) {
+      FINE.atMostEvery(1, SECONDS).log("Iteration %d", i)
```
gkossakowski, line 31: 💯rate limiting logging

gkossakowski:
this looks great

PR #262, opened 2018-11-10
[WIP] Kahan Summation in Translator

mio-stripe:
Adding Kahan summation of Exprs in Translator.


mio-stripe:
So far, the SBC's fail as expected, but produce reasonable numbers.
I get GC overloads with even small numbers of observations and traces show this is generating _tons_ of code so plenty of work to do.

avibryant:
I pushed https://github.com/stripe/rainier/tree/avi/ks-in-ir which merges in the current develop (to get the graphviz stuff) and makes a change that I think fixes the excess code gen without screwing up correctness, but I haven't done any real testing so please verify.

rainier-core/src/main/scala/com/stripe/rainier/compute/Translator.scala
```diff
@@ -171,11 +171,27 @@ private class Translator {
     combineTree(lazyExprs, ring)
   }
 
+  private def kahanSumUpdate(accumulator: Expr,
+                             compensation: Expr,
+                             term: Expr) = {
+    val nextExpr = binaryExpr(term, compensation, SubtractOp)
+    val newAccumulator = binaryExpr(accumulator, nextExpr, AddOp)
```
mio-stripe: With `ref(accumulator)` here, I get an uninitialized register error. As is, even with these other `refs` I still get hanging GC.

avibryant: Have you done graphviz plots to visualize what's going on?

mio-stripe:
So it turns out the GC overhead exceeded is because Packer is doing some way-too-deep recursion so this is not the way to go. Closing.

PR #264, opened 2018-11-13
fix lazy vals in benchmarks

avibryant:
I foolishly was using `lazy val` for benchmark fixtures, which had unpredictable results in what computation was actually being benchmarked. This cleans things up to properly use jmh `@setup` annotations, and also makes sure that the compilation benchmark is doing the work instead of potentially retrieving something cached.


mio-stripe:
LGTM

PR #266, opened 2018-11-14
Update tour.md

mio-stripe:
The main differences are that I removed the comparison of the log densities of Poisson(9) and Poisson(10) fit to data because showing how to compute the density seemed too involved. Happy to put it back. I also updated the hierarchical model at the end which feels a little involved because it uses a `Predictor.from` (not one of the existing helpers), so I'm open to suggestions there.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
@@ -56,6 +56,10 @@ object Predictor {
   def fromInt[Y, B](fn: Real => Distribution.Aux[Y, B]): Predictor[Int, Y] =
     from[Int, Y, Real, B](fn)
 
+  def fromIntPair[Y, B](
+      fn: ((Real, Real)) => Distribution.Aux[Y, B]): Predictor[(Int, Int), Y] =
+    from[(Int, Int), Y, (Real, Real), B](fn)
+
   def fromDouble[Y, B](
       fn: Real => Distribution.Aux[Y, B]): Predictor[Double, Y] =
     from[Double, Y, Real, B](fn)
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
@@ -127,6 +127,25 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       def gradient(index: Int) = outputs(index + 1)
     }
 
+  def densityValue(values: Map[Variable, Double] = Map.empty): Real = {
+    val rvVars = targetGroup.variables.sortBy(_.toString)
+    val (inputVars, inputVals) = values.toList.sortBy(_._1.toString).unzip
+    (rvVars, inputVars) match {
+      case (Nil, _) => targetGroup.base
+      case (rVs, iVs) if rVs == iVs =>
+        val den = density()
+        den.update(inputVals.toArray)
+        den.density
+      case _ => 0.0
+    }
+  }
+
+  def densityValue(value: Double): Real =
+    targetGroup.variables.headOption match {
+      case Some(v) => densityValue(Map(v -> value))
+      case _       => 0.0
+    }
+
   //this is really just here to allow destructuring in for{}
   def withFilter(fn: T => Boolean): RandomVariable[T] =
     if (fn(value))
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
@@ -127,24 +127,7 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       def gradient(index: Int) = outputs(index + 1)
     }
 
-  def densityValue(values: Map[Variable, Double] = Map.empty): Real = {
-    val rvVars = targetGroup.variables.sortBy(_.toString)
-    val (inputVars, inputVals) = values.toList.sortBy(_._1.toString).unzip
-    (rvVars, inputVars) match {
-      case (Nil, _) => targetGroup.base
-      case (rVs, iVs) if rVs == iVs =>
-        val den = density()
-        den.update(inputVals.toArray)
-        den.density
-      case _ => 0.0
-    }
-  }
-
-  def densityValue(value: Double): Real =
-    targetGroup.variables.headOption match {
-      case Some(v) => densityValue(Map(v -> value))
-      case _       => 0.0
-    }
+  lazy val densityValue: Real = targetGroup.base
 
   //this is really just here to allow destructuring in for{}
   def withFilter(fn: T => Boolean): RandomVariable[T] =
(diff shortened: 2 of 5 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -56,6 +56,10 @@ object Predictor {
   def fromInt[Y, B](fn: Real => Distribution.Aux[Y, B]): Predictor[Int, Y] =
     from[Int, Y, Real, B](fn)
 
+  def fromIntPair[Y, B](
```
avibryant, line 59: Little worried about combinatorial explosion here but obviously useful for now.

rainier-docs/src/main/tut/tour.md
```diff
@@ -10,39 +10,41 @@ import com.stripe.rainier.repl._
...
-
-You construct a distribution from its parameters, usually with the `apply` method on the object representing its family. So for example, this is a normal distribution with a mean of 0 and a standard deviation of 1:
+Rainier implements various familiar families of probability distributions like the [Normal](https://en.wikipedia.org/wiki/Normal_distribution) distribution, the [Uniform][UC] distribution, and the [Poisson](https://en.wikipedia.org/wiki/Poisson_distribution) distribution. You will find these three in [Continuous.scala](/rainier-core/src/main/scala/com/stripe/rainier/core/Continuous.scala) and [Discrete.scala](/rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala) - along with a few more, and we'll keep adding them as the need arises. You construct a distribution from its parameters, usually with the `apply` method on the object representing its family. So for example, this is a normal distribution with a mean of 0 and a standard deviation of 1:
```
avibryant, line 29: Purely out of curiosity, why the inconsistent application of footnote style links?

rainier-docs/src/main/tut/tour.md
```diff
@@ -103,21 +105,22 @@ val sales = List(4, 4, 7, 11, 8, 12, 10)
...
 We can model the number of sales we get on each day as a poisson distribution parameterized by the underlying rate. For example, looking at that data, we might guess that it came from a `Poisson(9)` or `Poisson(10)`. We can test those hypotheses by using the `fit` method available on all `Distribution` objects. Since `Poisson` is a `Distribution[Int]`, its `fit` will accept either `Int` or `Seq[Int]`, and return a `RandomVariable` which encodes the likelihood of the provided data being produced by that distribution. (We're not going to worry right now about what *kind* of `RandomVariable` it is, since we're only really interested in the likelihood).
 
-```tut
+```tut:silent
 val poisson9: RandomVariable[_] = Poisson(9).fit(sales)
 val poisson10: RandomVariable[_] = Poisson(10).fit(sales)
 ```
 
-Although it's almost never necessary, we can reach into any `RandomVariable` and get its current probability `density`:
+A `RandomVariable`'s density is a rather opaque object and, since it is almost never necessary to reach into a `RandomVariable` and get its current probability `density`, Rainier makes this hard to do. Taking on faith momentarily that
```
avibryant: Hmmmm. I think the "taking on faith approach" isn't worth it. What if we add something like `densityValue(values: Map[Variable,Double] = Map.empty)` that we could use here?

rainier-docs/src/main/tut/tour.md
```diff
@@ -277,24 +280,39 @@ Finally, let's close with a small and somewhat contrived example of a hierarchic
...
+
+or we can use Rainier's `Lookup` to package the two slopes into a single `RandomVariable` from which we can lookup the one we want. This means that our predictor now has to take both the index of the model to which the data belongs as well as the actual observation as input.
+
+```tut
+val allData = data.map{ case (x,y) => ((0, x), y) } ++ data2.map{ case (x, y) => ((1, x), y) }
+val regr4 = for {
+  slopes <- RandomVariable.fill(2)(LogNormal(0,1).param).map(Lookup(_))
+  intercept <- LogNormal(0,1).param
+  _ <- Predictor.fromIntPair{ case (index, x) => Poisson(intercept + slopes(index) * x) }.fit(allData)
+} yield (slopes(0), intercept)
```
avibryant, line 303: This is much better, thank you..

avibryant (commented):
The main remaining sticking point is the density() stuff. See my suggestion which shouldn't be too hard? And might be useful in other places.

rainier-docs/src/main/tut/tour.md
mio-stripe: One link had parens in it which broke when doing it inline. I don't remember why I moved some others down there :)

rainier-docs/src/main/tut/tour.md
mio-stripe: I'll try that

rainier-docs/src/main/tut/tour.md
mio-stripe: 👍

rainier-docs/src/main/tut/tour.md
mio-stripe: @avi-stripe I pushed something, but since for the docs the point is just that you get either a Constant or something opaque, I wonder if densityValue should just return `targetGroup.base`.

mio-stripe: Okay, decided to just return the `base` for now.

rainier-docs/src/main/tut/tour.md
```diff
@@ -103,21 +105,22 @@ val sales = List(4, 4, 7, 11, 8, 12, 10)
 
 We can model the number of sales we get on each day as a poisson distribution parameterized by the underlying rate. For example, looking at that data, we might guess that it came from a `Poisson(9)` or `Poisson(10)`. We can test those hypotheses by using the `fit` method available on all `Distribution` objects. Since `Poisson` is a `Distribution[Int]`, its `fit` will accept either `Int` or `Seq[Int]`, and return a `RandomVariable` which encodes the likelihood of the provided data being produced by that distribution. (We're not going to worry right now about what *kind* of `RandomVariable` it is, since we're only really interested in the likelihood).
 
-```tut
+```tut:silent
 val poisson9: RandomVariable[_] = Poisson(9).fit(sales)
 val poisson10: RandomVariable[_] = Poisson(10).fit(sales)
 ```
 
 Although it's almost never necessary, we can reach into any `RandomVariable` and get its current probability `density`:
```
avibryant: update to `densityValue`

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -127,6 +127,8 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       def gradient(index: Int) = outputs(index + 1)
     }
 
+  lazy val densityValue: Real = targetGroup.base
```
avibryant, line 130: This will be incorrect if there are placeholders but it's fine for now...

rainier-docs/src/main/tut/tour.md
mio-stripe: 👍

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
mio-stripe: 👍

avibryant:
LGTM

PR #268, opened 2018-11-20
fixes to mappings and placeholders

avibryant:
This changes Placeholder.variables to work like Placeholder.extract, with an explicit accumulator. This makes it easy to verify that the two implementations align properly (eg that ph.variables.zip(ph.extract) will match up variables with values correctly). Less importantly, it should also improve performance.

This PR also fixes a bug where the Map mapping had unstable key ordering which also contributed to misalignment of those two methods.


mio-stripe:
LGTM

PR #270, opened 2018-11-30
Refactor Likelihood and Placeholder

avibryant:
This adjusts the various types related to `Distribution` in order to rationalize and decouple them.
In this new world, we have:

* `Generator[T]`, as before, produces a posterior-predictive stream of `T` given a `RNG` and a `Numeric[Real]`
* `Likelihood[T]` now embodies a specific instance of a likelihood function: it has a `real` representation of the function, a `variables` list of the variable nodes that correspond to its observational input, and an `extract(t: T): Seq[Double]` method that takes an instance of `T` and encodes it as a double vector corresponding to the variable nodes.

We then have two typeclasses that allow implicit conversion of other objects into `Generator[T]` and `Likelihood[T]`:

* `ToGenerator[V,T]` will convert a `V` into a `Generator[T]`. For example, there's a `ToGenerator[Real,Double]`.
* `ToLikelihood[V,T]` will convert a `V` into a `Likelihood[T]`. For example, there's a `ToLikelihood[Discrete, Int]`.

Instead of `Placeholder[X,R]` we now have `Encoder[X,R]` which plays the same role of lifting an `X` value into `R` Real space, and which is generally a building block for `Likelihood` instances.

This lets us have a `Predictor[X,V]` which wraps a function from `R => V` whenever we have a `Encoder[X,R]`, and will provide a `fit((X,Y))` operation assuming we have a `ToLikelihood[V,Y]` [0], and will also provide a `generate(X): Y` operation assuming we have a `ToGenerator[V,Y]`; but there's no strict requirement to have either.

We can also now also create raw `Likelihood` instances with an API very close to that for `Predictor`, but where they wrap a function from `R => Real` that directly computes the likelihood, rather than going through an existing distribution or other likelihood function.

Note that we can also automatically derive a `ToLikelihood[Predictor[X,V],(X,Y)]`, which means that we can nest one `Predictor` inside another, or nest a `Likelihood` inside a `Predictor`.

Part of this refactoring is removing `Wrapping` and `Mapping` and completing separating (once again) the idea of encoding stuff *into* real space and generating stuff *out* of real space. One impact is that we've lost the ability (for now) to have observations that are `Map[K,V]`; this will be replaced by something that makes it easy to use case classes for observations instead, in some future PR.


rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala
```diff
@@ -4,42 +4,68 @@ import com.stripe.rainier.compute._
...
-      val doubles = ph.extract(t, Nil)
+      val doubles = extract(t)
       arrayBufs.zip(doubles).foreach {
         case (a, d) => a += d
       }
     }
     val placeholdersMap =
-      variables.zip(arrayBufs.map(_.toArray)).toMap
-    new Target(real, placeholdersMap)
+      placeholders.zip(arrayBufs.map(_.toArray)).toMap
```
mio-stripe, line 23: I'm on board with these pretty minor changes, but I still don't totally understand why this works this way. I'm imagining that T is (Double, Double) so for a given pair of Doubles, placeholders is a list of two variables and extract gives us the list with those two doubles. Then, given a seq of pairs of doubles, `arrayBufs` is a list of two ArrayBuffers and for each pair in our list, we extract the list of Doubles and zip and add, then put that in placeholdersMap which goes into a new RandomVariables's `targets` so morally these terms are getting added to the density. You mentioned that this is a chunk of the compute graph. Is the idea that some other chunk might also fit to the same observations? Or are the observations supposed to get spread around?
Also, my understanding is that now, `Likelihoods` are only used for computing likelihoods in MCMC and are totally decoupled from Distributions in general so you can make whatever likelihood you want, but can also use `Distribution.likelihood` to make a `Likelihood` out of a `Distribution`. That's nice.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -5,66 +5,65 @@ import com.stripe.rainier.compute._
...
-    new Predictor[X, Y] {
-      type Q = A
-      type R = B
-
-      val xq = xa
-      def create(q: Q) = fn(q)
+object Predictor extends LikelihoodMaker {
+  def likelihood[L, X, Y](pred: Predictor[L, X])(
+      implicit lh: ToLikelihood[L, Y]): Likelihood[(X, Y)] = {
+    val (p, v) = pred.xp.create(Nil)
```
mio-stripe: Nit: if `v` were `vs` or `variables` or `placeholders` it would be a reminder that they are a list and what they are.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -5,66 +5,65 @@ import com.stripe.rainier.compute._
...
   */
-trait Predictor[X, Y] extends Likelihood[(X, Y)] {
-  private[core] type Q
-  private[core] type R
-  type P = (Q, R)
-  private[core] def create(q: Q): Distribution.Aux[Y, R]
-  private[core] def xq: Wrapping[X, Q]
+sealed trait Predictor[L, X] { self =>
+  type P
+  protected def xp: Encoder[X, P]
```
mio-stripe: Nit: maybe `encoder` rather than `xp`

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -5,66 +5,65 @@ import com.stripe.rainier.compute._
 /**
   * Predictor class, for fitting data with covariates
   */
-trait Predictor[X, Y] extends Likelihood[(X, Y)] {
-  private[core] type Q
-  private[core] type R
-  type P = (Q, R)
-  private[core] def create(q: Q): Distribution.Aux[Y, R]
-  private[core] def xq: Wrapping[X, Q]
+sealed trait Predictor[L, X] { self =>
```
mio-stripe: This is a huge nit, but since this contains a type `P`, an `Encoder[X, P]` and a function `P => L`, it would be less confusing for it to be a `Predictor[X, L]`.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -5,66 +5,65 @@ import com.stripe.rainier.compute._
...
+      val placeholders = v ++ inner.placeholders
+      def extract(t: (X, Y)) =
+        pred.xp.extract(t._1, Nil) ++ inner.extract(t._2)
     }
+  }
 
-  def fromInt[Y, B](fn: Real => Distribution.Aux[Y, B]): Predictor[Int, Y] =
-    from[Int, Y, Real, B](fn)
+  class Maker[X, A](xa: Encoder[X, A]) {
+    def apply[B](fn: A => B): Predictor[B, X] =
```
mio-stripe: Changing the order of the type variables in `Predictor` would mean that a `Maker[X, A]` takes an `Encoder[X, A]` and a function `A => B` and produces a `Predictor[X, B]` which seems less jarring.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -5,66 +5,65 @@ import com.stripe.rainier.compute._
...
+        type P = A
+        val xp = xa
+        def create(p: P) = fn(p)
+      }
+  }
 
-  def fromIntPair[Y, B](
-      fn: ((Real, Real)) => Distribution.Aux[Y, B]): Predictor[(Int, Int), Y] =
-    from[(Int, Int), Y, (Real, Real), B](fn)
+  type M[X, A] = Maker[X, A]
```
mio-stripe: I don't this gets used.

rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala
```diff
@@ -4,42 +4,68 @@ import com.stripe.rainier.compute._
...
+  def fromIntPair = maker[(Int, Int), (Real, Real)]
+  def fromDoublePair = maker[(Double, Double), (Real, Real)]
+  def fromIntVector(size: Int) =
+    maker[Seq[Int], Seq[Real]](Encoder.vector(size))
+  def fromDoubleVector(size: Int) =
+    maker[Seq[Double], Seq[Real]](Encoder.vector(size))
+}
+
+object Likelihood extends LikelihoodMaker {
+  class Maker[T, P](ph: Encoder[T, P]) {
```
mio-stripe: I don't totally see what these Maker and Fitter classes are doing and it's a little confusing to have slightly different classes with the same names in the Likelihood and Predictor objects. Could these be functions? Or implicit classes where `Maker[T, P]` wraps an `Encoder[T, P]` and has `toLikelihood` and `toPredictor` methods and `Fitter[T, P]` wraps an `Encoder[T,P]` that return a `RandomVariable[Unit]` or a `RandomVariable[Predictor[_,_]]` given the right things?

rainier-core/src/main/scala/com/stripe/rainier/compute/Encoder.scala
```diff
@@ -0,0 +1,60 @@
...
+      def create(acc: List[Variable]) = {
+        val (bv, acc1) = b.create(acc)
+        val (av, acc2) = a.create(acc1)
+        ((av, bv), acc2)
+      }
+      def extract(t: (A, B), acc: List[Double]) =
+        a.extract(t._1, b.extract(t._2, acc))
+    }
+
+  def vector[T, U](size: Int)(
```
mikeheaton-stripe: Should this be an `implicit def vector`?

avibryant: It can't be because it can't implicitly know the size, so you have to tell it that.

We could also/alternatively do a `ToVector` kinda like your `ToMap` to solve this. So eg maybe we could have you do:

 ```scala
case class MyDataVec(vec: Seq[Double])
object MyDataVec {
 implicit val toVec = new ToVector[MyDataVec] {
    val size = 10
    def apply(v: MyDataVec) = v.vec
}
```

and then we could have an implicit for that.

File: rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala (hunk 6 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala]: It does so that both `Predictor` and `Likelihood` can inherit from `LikelihoodMaker`. (I agree with you that this is all crazy convoluted).

File: rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala (hunk 7 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala]: Hmmm. I'll see what I can think of here to simplify and make more clear. The scala type inferencer is a very picky customer.

File: rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala (hunk 5 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala]: Will do.

File: rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala (hunk 4 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala]: Yep will do.

File: rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala (hunk 3 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala]: On it.

File: rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala (hunk 2 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala]: good call

File: rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala (hunk 1 above)
Comment (avibryant) [reply rainier-core/src/main/scala/com/stripe/rainier/core/Likelihood.scala:23]: > You mentioned that this is a chunk of the compute graph. Is the idea that some other chunk might also fit to the same observations? Or are the observations supposed to get spread around?

It's a chunk of the compute graph in that it represents the contribution to the overall joint density of this single likelihood function; but there are also all the priors on the parameters that contribute, and then there may also be other observations, with some different structure, that get fit separately.

Does that answer your question?

File: rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala [hunk 9]
```diff
@@ -5,66 +5,66 @@ import com.stripe.rainier.compute._
...
-    from[Double, Y, Real, B](fn)
+  trait From[X, U] {
+    def from[L](fn: U => L): Predictor[X, L]
+    def fromVector[L](k: Int)(fn: IndexedSeq[U] => L): Predictor[Seq[X], L]
+  }
 
-  def fromDoubleVector[Y, B](
-      fn: Seq[Real] => Distribution.Aux[Y, B]): Predictor[Seq[Double], Y] =
-    from[Seq[Double], Y, Seq[Real], B](fn)
+  def apply[X](implicit enc: Encoder[X]) =
```
gkossakowski, line 53: your life will be much better if you annotate return types of public methods _especially_ when path-dependent types are involved

this is a general comment that i placed here since i was looking at the shape of the code that run into the compiler crash

rainier-core/src/main/scala/com/stripe/rainier/compute/Encoder.scala
```diff
@@ -0,0 +1,70 @@
...
+trait Encoder[T] {
+  type U
+  def wrap(t: T): U
+  def create(acc: List[Variable]): (U, List[Variable])
+  def extract(t: T, acc: List[Double]): List[Double]
+}
+
+object Encoder {
+  type Aux[X, Y] = Encoder[X] { type U = Y }
+  implicit val int: Aux[Int, Real] =
```
mio-stripe, line 12: Does the `: Aux[Int, Real]` help with type inference in other places? Is there something weird about that since `U` is a type member, the compiler doesn't know both `Int` and `U=Real` at the same time without it?

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -5,66 +5,66 @@ import com.stripe.rainier.compute._
...
   */
-trait Predictor[X, Y] extends Likelihood[(X, Y)] {
-  private[core] type Q
-  private[core] type R
-  type P = (Q, R)
-  private[core] def create(q: Q): Distribution.Aux[Y, R]
-  private[core] def xq: Wrapping[X, Q]
+sealed trait Predictor[X, L] { self =>
+  type P
+  protected def encoder: Encoder[X] { type U = P }
```
mio-stripe, line 10: Would it be equivalent to annotate this as `Encoder.Aux[X, P]`?

rainier-docs/src/main/tut/tour.md
```diff
@@ -243,7 +243,7 @@ Like `Distribution`, `Predictor` implements `fit` (in fact, they both extend the
 ```tut
 val regr = for {
     (slope, intercept) <- prior
-    predictor <- Predictor.fromInt{i =>
+    predictor <- Predictor[Int].from{i =>
```
mio-stripe, line 246: Nice

mio-stripe (approved):
Looks great! Left a few non-blocking questions about `Encoder.Aux`. 🚢

rainier-core/src/main/scala/com/stripe/rainier/compute/Encoder.scala
avibryant: No, it doesn't affect anything other than how concisely I can express the annotations.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
avibryant: Yes.

PR #272, opened 2018-12-10
Predictor.lookup

avibryant:
This makes it easier to use `Lookup` for hierarchical models and similar. If you have your data as a `Seq[(K,V)]`, where `K` indicates some kind of segmentation, you can now create a `Map[K,Real]` of parameters, and then build a predictor like this:

```scala
val params: Map[K,Real] = ???
val data: Seq[(K,V)] = ???

Predictor.lookup(params){param => ...}.fit(data)
```

where the `...` creates some kind of `L` with `ToLikelihood[L,V]`, ie, most likely a `Likelihood` object or a nested `Predictor`.


mio-stripe:
LGTM

PR #273, opened 2018-12-11
stack safe Packer

avi-stripe:
This uses the scala stdlib's trampolining to make Packer stack safe. This seems to have been the main stack-space bottleneck for deeply recursive models like DLM - without this change, it blows up above 15 observations. After this change, we can get into large enough numbers of iterations that it falls over for other reasons :)

cc @jonnylaw

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/ir/Packer.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/ir/Packer.scala
@@ -1,52 +1,64 @@
 package com.stripe.rainier.ir
 
+import scala.util.control.TailCalls
+import TailCalls.TailRec
+
 private class Packer(methodSizeLimit: Int) {
   private var methodDefs: List[MethodDef] = Nil
   def methods = methodDefs
 
   def pack(p: Expr): MethodRef = {
-    val (pExpr, _) = traverse(p, 0)
+    val (pExpr, _) = traverse(p, 0).result
     createMethod(UnaryIR(pExpr, NoOp))
   }
 
-  private def traverse(p: Expr, parentSize: Int): (Expr, Int) =
+  private def traverse(p: Expr, parentSize: Int): TailRec[(Expr, Int)] =
     p match {
       case v: VarDef => traverseVarDef(v, parentSize)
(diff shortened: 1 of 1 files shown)


rainier-core/src/main/scala/com/stripe/rainier/ir/Packer.scala
```diff
@@ -1,52 +1,64 @@
...
-  private def traverseIR(p: IR): (IR, Int) =
+  private def traverseIR(p: IR): TailRec[(IR, Int)] =
     p match {
       case b: BinaryIR =>
-        val (leftExpr, leftSize) =
-          traverse(b.left, 2)
-        val (rightExpr, rightSize) =
-          traverse(b.right, leftSize + 1)
-        (new BinaryIR(leftExpr, rightExpr, b.op), leftSize + rightSize + 1)
+        traverse(b.left, 2).flatMap {
```
mio-stripe, line 36: This `flatmap...map` could equally be

```
for {
  (leftExpr, leftSize) <- traverse(b.left, 2)
  (rightExpr, rightSize) <- traverse(b.right, leftSize + 1)
} yield (new BinaryIR(leftExpr, rightExpr, b.op), leftSize + rightSize + 1)
```

I was going to call it six of one, a half dozen of the other, but I think this way is a bit easier to parse.

rainier-core/src/main/scala/com/stripe/rainier/ir/Packer.scala
```diff
@@ -1,52 +1,64 @@
...
   def methods = methodDefs
 
   def pack(p: Expr): MethodRef = {
-    val (pExpr, _) = traverse(p, 0)
+    val (pExpr, _) = traverse(p, 0).result
     createMethod(UnaryIR(pExpr, NoOp))
   }
 
-  private def traverse(p: Expr, parentSize: Int): (Expr, Int) =
+  private def traverse(p: Expr, parentSize: Int): TailRec[(Expr, Int)] =
```
mio-stripe, line 15: This is excellent!

mio-stripe (approved):
Awesome. Made a non-blocking readability suggestion.

PR #274, opened 2018-12-12
EHMC for tuning free HMC

jonnylaw:
I've started writing the implementation of EHMC (https://arxiv.org/pdf/1810.04449.pdf) which aims to tune the step-size and number of leapfrog steps using Dual averaging and by constructing an empirical distribution consisting of the number of steps until a u-turn. This algorithm is more straightforward to implement than NUTS and performs better according to the arxiv paper.

In order to re-use some of the logic from LeapFrog I had to expose two private methods and add a new method which performs L leapfrogs and returns a value of the parameters and momentum - this allows us to check the u-turn criteria while continuing in a single direction. 

I'm out of practice with imperative programming - so something could have gone wrong in the implementation!


rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,26 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
 
+  /**
+    * Perform l leapfrog steps starting at position q and momentum p
+    * @param l the total number of leapfrog steps
+    * @param stepSize the current step size
+    * @param pq an array containing the momentum, parameters and log-density
+    * @return the new value of the parameters and momentum
+    * array with updated density
+    */
+  def leapfrogs(l: Int, stepSize: Double, pq: Array[Double]): Array[Double] = {
```
avi-stripe: maybe call this `steps`? And you should use it from `tryStepping` and `step`

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
...
+          iterations: Int,
+          lf: LeapFrog,
+          params: Array[Double])(implicit rng: RNG): Vector[Int] = {
+
+    Vector
+      .iterate((params, l0), iterations)(longestBatchStep(l0, eps, lf))
+      .map(_._2)
+  }
+
+  def uturnCondition(theta: Array[Double],
```
avi-stripe: If you moved this into `LeapFrog` it could maybe be `isUTurn(start: Array[Double], finish: Array[Double])`? Then you could get away without the allocation of new arrays in `variables()` and `momentum()`.

(Also, I find the explicit indication of which way the boolean goes better than just calling it `condition`).

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,26 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+    copy(pq, pqBuf)
+    initialHalfThenFullStep(stepSize)
+    var i = 1
+    while (i < l) {
+      twoFullSteps(stepSize)
+      i += 1
+    }
+    finalHalfStep(stepSize)
+    pqBuf
```
avi-stripe: I'm uncomfortable with returning pqBuf to users outside this class. Maybe this should be a private method that returns Unit, and there can be a public version that copies pqBuf into an output buffer at the end?

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
```diff
@@ -0,0 +1,41 @@
+package com.stripe.rainier.sampler
+
+import scala.collection.mutable.ListBuffer
+
+/**
+  * Empirical HMC - automiating tuning of step size and number of leapfrog steps
```
avi-stripe: typo: "automating"

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
```diff
@@ -0,0 +1,41 @@
...
+  * @param l0 the initial number of leapfrog steps to use during the dual averaging phase
+  * @param k the number of iterations to use when determining the
+  * empirical distribution of the total number of leapfrog steps until a u-turn
```
avi-stripe, line 9: Do `l0` and `k` have good defaults? How common is it to need to tune them?

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
```diff
@@ -0,0 +1,41 @@
...
+    val empiricalL = LongestBatch.run(l0, stepSize, k, lf, params)
+    if (stepSize == 0.0)
+      List(lf.variables(params))
+    else {
+      val buf = new ListBuffer[Array[Double]]
+      var i = 0
+
+      while (i < iterations) {
+
+        val j = LongestBatch.discreteUniform(0, k)
```
avi-stripe: if `LongestBatch.run` returned a `Generator[Int]` via `Categorical.list(vec).generator`, then this could just be `nSteps = empiricalL.get`.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+      (params, l)
+    }
+  }
+
+  /**
+    * Calculate a vector representing the empirical distribution
+    * of the steps taken until a u-turn
+    */
+  def run(l0: Int,
```
avi-stripe: as noted earlier, this could return a `Generator[Int]`. Also, `run` is a bit generic, maybe `findEmpiricalDistribution`?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
+package com.stripe.rainier.sampler
+
+/**
+  * Compute a distribution of number of leapfrog steps required to make
+  * a u-turn using a tuned value of the step size eps
+  * @param l0 the initial number of leapfrog steps
+  * @param eps the tuned leapfrog step size
+  */
+object LongestBatch {
```
avi-stripe: what's the significance of `Batch` here? A bit unclear to me.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
...
+    * of the steps taken until a u-turn
+    */
+  def run(l0: Int,
+          eps: Double,
+          iterations: Int,
+          lf: LeapFrog,
+          params: Array[Double])(implicit rng: RNG): Vector[Int] = {
+
+    Vector
+      .iterate((params, l0), iterations)(longestBatchStep(l0, eps, lf))
```
avi-stripe: I find the currying here a bit opaque, could you flesh this out to be a bit more explicit (and maybe with only one arg list for longestBatchStep?). Also if you pass in `params` to be modified here, you probably just need a `map` anyway.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
...
+      i += 1
+    }
+
+    if (out.isNaN)
+      false
+    else
+      out >= 0
+  }
+
+  def discreteUniform(min: Int, max: Int)(implicit rng: RNG) = {
```
avi-stripe: feels like the wrong place for this; suggest using `Categorical.list` but alternatively you could add a `DiscreteUniform` distribution and use that.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
avi-stripe: I'd suggest moving `longestBatch`, `longestBatchStep`, and `uturnCondition` into `LeapFrog`, and maybe moving `run` into `EHMC`.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
...
+    * Perform a single step of the longest batch step algorithm
+    */
+  def longestBatchStep(l0: Int, eps: Double, lf: LeapFrog)(
+      thetal: (Array[Double], Int))(implicit rng: RNG): (Array[Double], Int) = {
+
+    val params = thetal._1
+    lf.initializePs(params)
+    val (tp, l) = longestBatch(params, l0, eps, lf)
+    val prop = if (l < l0) {
+      lf.leapfrogs(l0 - l, eps, tp)
```
avi-stripe: Once in `LeapFrog`, you should model this off `step`, in terms of copying params into pqBuf and then having `leapfrogs` modify that - rather than returning a new array.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
```diff
@@ -0,0 +1,97 @@
...
+  * a u-turn using a tuned value of the step size eps
+  * @param l0 the initial number of leapfrog steps
+  * @param eps the tuned leapfrog step size
+  */
+object LongestBatch {
+
+  /**
+    * Calculate the longest-step size until a u-turn
+    */
+  def longestBatch(params: Array[Double],
```
avi-stripe: would it be accurate to name this `stepUntilUTurn`?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
avi-stripe: In the case where `l < l0`, then `tp` will be the same as `params` (because of how `longestBatch` works. Does that mean you'd want to advance `l0` steps instead of `l0 - l`?

In general, it looks to me like we're trying to do two separate things that are weirdly entangled:
- advance exactly `l0` steps and record the param values at that point
- discover how many steps we need until a u-turn

Is that correct? I wonder if we can separate those two concerns out more cleanly without wasting any computation. It seems like it would be good to, if we can, whereas right now they're both kinda smeared across this method and `longestBatch`.

avi-stripe (commented):
Thank you so much for working on this! This is great. 

I've left a bunch of comments. Some of them are high level and some of them are more detailed, and (maybe confusingly) some of the detailed ones are probably rendered moot by the changes required to address the high level ones; it was hard to do a more coherent set of suggestions without it turning into a rewrite.

Also: to what extent do you think that, once we have this, we even need the old `HMC`? Should we just standardize on this and drop the other implementation?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
jonnylaw: I have renamed this steps, makes more sense! I've changed `step` to use `steps`

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
jonnylaw: Agreed.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
jonnylaw: No problem, I altered this.

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
jonnylaw: There isn't any guidance in the paper for the value of l0, however it is suggested that a dual averaging scheme could be implemented to determine the initial value l0 for a given target acceptance rate. As for k, 2,000 iterations are used in all the examples in the paper so that's what I've been using.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
jonnylaw: The algorithm is called longest batch in the paper so I just called the class the same thing, it actually refers to the longest step size in the same direction before a u-turn. I've moved these functions to Leapfrog and removed the class.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
jonnylaw: I've added a bit more, but I haven't replaced vector.iterate with an explicit loop.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
jonnylaw: Yes I thought that, I'm not sure how best to use the Categorical distribution here. If I create a categorical distribution with equal probabilities:

`val cat = Categorical.list(Seq.fill(k)(1.0 / k))`

Then what would be the best way to sample from it? I can call `generator` on it, then call `get` but I would need an evaluator and I'm not sure what the evaluator would look like in this case.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
jonnylaw: That's exactly what we're trying to do, but without duplicating computations. I'll have a think about how exactly to separate this.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LongestBatch.scala
jonnylaw: Yep, I just used the function name from the paper.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
     pqBuf(potentialIndex) = density.density * -1
   }
 
+  /**
+    * Perform l leapfrog steps starting at position q and momentum p
+    * @param l the total number of leapfrog steps to perform
+    * @return the new value of the parameters and momentum
```
avi-stripe: this part of the comment is no longer accurate

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+    initialHalfThenFullStep(stepSize)
+    var i = 1
+    while (i < l) {
+      twoFullSteps(stepSize)
+      i += 1
+    }
+    finalHalfStep(stepSize)
+  }
+
+  private def isUTurn(theta: Array[Double], pqP: Array[Double]): Boolean = {
```
avi-stripe: Given that the other methods in the call chain just directly use `pqBuf`, is there any reason to pass it in as an arg here?

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+    l
+  }
+
+  /**
+    * Perform a single step of the longest batch step algorithm
+    */
+  private def longestBatchStep(l0: Int, stepSize: Double)(
+      implicit rng: RNG): Int = {
+
+    val params = pqBuf
```
avi-stripe: This is just aliasing the pqBuf array, and will track any modifications to it, which means your `logAcceptanceProb` call is broken (they will be the same array). I suggest you pass in a separate `params` array and `copy(params, pqBuf)` here (after first initializing it, I guess).

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+  }
+
+  /**
+    * Calculate a vector representing the empirical distribution
+    * of the steps taken until a u-turn
+    */
+  def empiricalLongestSteps(l0: Int, k: Int, stepSize: Double)(
+      implicit rng: RNG): Vector[Int] = {
+
+    Vector.fill(k)(longestBatchStep(l0, stepSize))
```
avi-stripe: Per the note in `EHMC`, I suspect you want to be passing in `params` here, and either passing it directly into `longestBatchStep`, or cloning it once and then using that clone for every call to `longestBatchStep`.

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
```diff
@@ -0,0 +1,50 @@
...
+             warmupIterations: Int,
+             iterations: Int,
+             keepEvery: Int)(implicit rng: RNG): List[Array[Double]] = {
+
+    val lf = LeapFrog(density)
+    val params = lf.initialize
+    val stepSize =
+      DualAvg.findStepSize(lf, params, 0.65, l0, warmupIterations)
+
+    val empiricalL: Vector[Int] = lf.empiricalLongestSteps(l0, k, stepSize)
```
avi-stripe: `findStepSize` acts as a warmup and advances the `params` vector. Two related questions:
* Do you want to be starting the empiricalLongestSteps trajectories from the vector that findStepSize ended with?
* Do you want to be using the empiricalLongestSteps work as additional warmup and continue the actual samples from the point it ends up at, or do you want to throw them away?

Either way you probably want to be passing in `params` here.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+
+    val initTheta = variables(pqBuf)
+    var out = pqBuf
+    var l = 0
+    while (!isUTurn(initTheta, pqBuf)) {
+      l += 1
+      steps(1, stepSize)
+
+      if (l == l0)
+        out = pqBuf
```
avi-stripe: This is just aliasing a mutable array. If you want to be snapshotting the state at the `l0`th step, so that you can rewind to it after keeping going until a u-turn, you need to be using `copy` into some other buffer. I suggest you allocate the buffer once in `empiricalLongestSteps` and thread it through to here.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+    * Perform a single step of the longest batch step algorithm
+    */
+  private def longestBatchStep(l0: Int, stepSize: Double)(
+      implicit rng: RNG): Int = {
+
+    val params = pqBuf
+    initializePs(params)
+    val l = longestStepUntilUTurn(l0, stepSize)
+    if (l < l0)
+      steps(l0 - l, stepSize)
```
avi-stripe: I would feel better if the contract of `longestStepUntilUTurn` was that it always advanced `pqBuf` exactly `l0` steps, ie, if this check and extra `steps` call were moved inside it.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +60,85 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+    else
+      out < 0
+  }
+
+  /**
+    * Calculate the longest-step size until a u-turn
+    */
+  private def longestStepUntilUTurn(l0: Int, stepSize: Double): Int = {
+
+    val initTheta = variables(pqBuf)
```
avi-stripe: `variables` allocates. It would be better if we reused a buffer here. (As per the note just below, we may need to allocate that in `empiricalLongestSteps` and thread it through).

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    if (stepSize == 0.0)
+      List(lf.variables(params))
+    else {
+      val buf = new ListBuffer[Array[Double]]
+      var i = 0
+
+      while (i < iterations) {
+
+        // val j: Int = Categorical.list(Seq.fill(k)(1.0 / k)).generator.get
```
avi-stripe: I've changed my mind about using `Categorical` here since that would introduce a dependency between `sampler` and `core` which I've been trying to avoid.

However I've just noticed that `RNG` already has an `int` method which you can use here: `val j = rng.int(k)`.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
jonnylaw: Ah yes, I'm not very good at reasoning with mutable state...

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -92,51 +96,66 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
-  def empiricalLongestSteps(l0: Int, k: Int, stepSize: Double)(
-      implicit rng: RNG): Vector[Int] = {
+  def empiricalLongestSteps(
+      params: Array[Double],
+      l0: Int,
+      k: Int,
+      stepSize: Double)(implicit rng: RNG): Vector[Int] = {
 
-    Vector.fill(k)(longestBatchStep(l0, stepSize))
+    var i = 0
```
avi-stripe: I actually think the `Vector.fill` was just fine here; either way you're allocating a new vector, and you know the size going in so ArrayBuffer doesn't buy you anything.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +61,103 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+                                    l0: Int,
+                                    stepSize: Double): Int = {
+
+    val out = params
+    var l = 0
+    while (!isUTurn(params)) {
+      l += 1
+      steps(1, stepSize)
+      if (l == l0)
+        copy(pqBuf, out)
```
avi-stripe: `out` is an alias of `params` so you're trashing your params array here which will mess up your acceptance prob computation later. You're going to need some other buffer you can save this version of pqBuf to. I think it's fine to have that buffer as an instance variable of `LeapFrog` that only gets used by this method (as long as you explain that in a comment).

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -60,6 +61,103 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
...
+    }
+    if (l < l0) {
+      steps(l0 - l, stepSize)
+      copy(pqBuf, out)
```
avi-stripe: Wasteful to `copy(pqBuf, out)` immediately followed by `copy(out, pqBuf)`. Would be better to just avoid both copies in the case that `l < l0`.

avi-stripe (dismissed):
Sorry for the long delay! There's one last mutable state issue to fix but it's nearly mergeable. Thanks so much.

avibryant:
See my minor nit comment above but otherwise I think this is ready to merge! Thank you so much for the contribution. Excited to see how this performs.

PR #275, opened 2018-12-14
add asMap encoder

mikeheaton-stripe:
Adds an `asMap` encoder, worked on with @avi-stripe and @mio-stripe. The encoder provides a way to encode a numeric case class, or other similar objects such as a `Map[String, Double]`-like with known keys, into Real-space by supplying a `ToMap` definition on the companion object.

Testing: Passes existing tests. Used in Stripe-internal repo. 

r? @avi-stripe @mio-stripe


avi-stripe:
LGTM!

PR #276, opened 2018-12-19
logging for HMC

avi-stripe:
Adding some logging to the HMC sampler:

* at `FINE`, logs the stages of warmup/sampling
* at `FINER`, logs individual iterations (no more than once per second)
* at `FINEST`, logs each iteration with accept/reject info


mikeheaton-stripe:
LGTM

PR #321, opened 2019-01-17
Fix compute bug with batched observations

avi-stripe:
This fixes a hard-to-trigger bug in the implementation of mini-batched likelihoods. The problem was:

* in `compileTargets`, we construct a single `CompiledFunction` that includes, as distinct outputs, `k+1` extra versions of the likelihood function; each one sums the likelihoods for `2^i` observations, for `i` in `0..k`.
* when compiling multiple outputs, they are treated by the compiler as being evaluated in order, and it is legal for output `n` to rely on intermediate computations stored in the `globals` array by any output `< n`.
* the `k+1` extra versions of the likelihood function were given to the compiler in ascending order from 0..k 
* however, in `DataFunction`, we were computing the outputs in reverse order, from k..0. This meant that in the case that one of these outputs depended on intermediate state from a previous output, that state will not have been initialized yet, and we'd get the wrong answer.
* this problem was rare because we deliberately include and use a separate, special version of the likelihood function for the very first observation, intending for that to compute any intermediate values that can be shared across outputs. The bug would only occur when, for some reason, there's some sharing that can occur between two of the other outputs but *not* with this one. (I'm still unclear on why that ever happens, but I accept that it's possible).

For now, we fix this simply by reversing the order of the outputs as we provide them to the compiler, so that it matches their evaluation order in `DataFunction`.

This PR also adds tests that would fail before this fix, and just might catch future regressions.

This PR does not fix the real underlying problem here, which is that `Target` and `DataFunction` are extremely brittle, tightly coupled, and confusing pieces of code. 

Thanks to @kailuowang for the initial bug report on this.

### DIFF
--- a/rainier-benchmark/src/main/scala/com/stripe/rainier/bench/SBCBenchmark.scala
+++ b/rainier-benchmark/src/main/scala/com/stripe/rainier/bench/SBCBenchmark.scala
@@ -41,7 +41,7 @@ abstract class SBCBenchmark {
 
   @Benchmark
   def compile() =
-    Compiler.default.compileTargets(TargetGroup(model.targets), true, 4)
+    Compiler.default.compileTargets(TargetGroup(model.targets, 0), true, 4)
 
   @Benchmark
   def run() =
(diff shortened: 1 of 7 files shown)


kailuowang:
Thanks so much @avi-stripe !

mio-stripe:
LGTM

PR #323, opened 2019-02-07
cleanups and logging for hmc

avi-stripe:
This does not change any behavior, but adds logging and some refactoring to the various HMC-related classes, to make it easier to debug sampler problems. I'll leave review comments on individual lines of code to explain.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/sampler/DualAvg.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/sampler/DualAvg.scala
@@ -7,7 +7,7 @@ final private class DualAvg(
     delta: Double,
     var logStepSize: Double,
     var logStepSizeBar: Double,
-    var avgAcceptanceProb: Double,
+    var avgError: Double,
     var iteration: Int,
     shrinkageTarget: Double,
     stepSizeUpdateDenom: Double = 0.05,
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/sampler/DualAvg.scala
```diff
@@ -7,7 +7,7 @@ final private class DualAvg(
     delta: Double,
     var logStepSize: Double,
     var logStepSizeBar: Double,
-    var avgAcceptanceProb: Double,
+    var avgError: Double,
```
avi-stripe, line 10: clarifying this value (Hbar in the paper): it's not the avg acceptance prob, it's the avg difference between the acceptance prob and delta.

rainier-core/src/main/scala/com/stripe/rainier/sampler/DualAvg.scala
```diff
@@ -50,38 +45,34 @@ private object DualAvg {
...
-  def findStepSize(lf: LeapFrog,
-                   params: Array[Double],
-                   delta: Double,
-                   nSteps: Int,
-                   iterations: Int)(implicit rng: RNG): Double = {
-    FINE.log("Finding reasonable initial step size")
-    val stepSize0 = findReasonableStepSize(lf, params)
-    FINE.log("Found initial step size of %f", stepSize0)
-
+  def findStepSize(delta: Double, stepSize0: Double, iterations: Int)(
```
avi-stripe, line 53: we can simplify this method quite a bit and make it much more flexible by separating out the function from stepSize => logAcceptanceProb, and letting the caller supply that. Similarly, we take an initial step size rather than hard coding the findReasonableStepSize call.

rainier-core/src/main/scala/com/stripe/rainier/sampler/Ehmc.scala
```diff
@@ -1,44 +1,76 @@
...
+
   def sample(density: DensityFunction,
              warmupIterations: Int,
              iterations: Int,
              keepEvery: Int)(implicit rng: RNG): List[Array[Double]] = {
 
     val lf = LeapFrog(density)
     val params = lf.initialize
+
+    FINE.log("Finding reasonable initial step size")
```
avi-stripe, line 23: we've pulled a lot more of the logic into this class and added logging. This becomes somewhat duplicative of HMC, but means it can be much more flexibly customized. For example, we might want to use longestBatchStep instead of step during the dual averaging.

rainier-core/src/main/scala/com/stripe/rainier/sampler/HMC.scala
```diff
@@ -25,16 +31,21 @@ final case class HMC(nSteps: Int) extends Sampler {
...
         i += 1
+        FINER
+          .atMostEvery(1, SECONDS)
+          .log("Sampling iteration %d of %d, acceptance rate %f",
+               i,
+               iterations,
+               (acceptSum / i))
       }
-      FINE.log("Finished sampling")
+      FINE.log("Finished sampling, acceptance rate %f", (acceptSum / i))
```
avi-stripe, line 48: So nice to have acceptance rate output.

rainier-core/src/main/scala/com/stripe/rainier/sampler/LeapFrog.scala
```diff
@@ -127,9 +132,8 @@ private[sampler] case class LeapFrog(density: DensityFunction) {
     * @param l0 the initial number of steps
     * @param stepSize the current value of the leapfrog step size
     */
-  private def longestBatchStep(params: Array[Double],
-                               l0: Int,
-                               stepSize: Double)(implicit rng: RNG): Int = {
+  def longestBatchStep(params: Array[Double], l0: Int, stepSize: Double)(
```
avi-stripe, line 135: This is now public and returns the acceptance prob so that it can be used, if desired, during dual averaging.

avi-stripe:
self-merging in advance of release

PR #326, opened 2019-02-11
writeCSV

avi-stripe:
This is a tiny little utility for writing out CSV files from `Seq[Map[String,Double]]`. The intended use is as a barebones arviz integration; if you write out your samples in this format, you can then do something like this:

screen shot 2019-02-10 at 8 22 07 pm


avibryant:
@mio-stripe bump

avibryant:
timing out and merging, feel free to leave post-merge review

PR #328, opened 2019-02-18
Changed Predictor.fromInt to Predictor[Int].from in README and tour

mfirry:
Hi,

while trying to use `rainier` It looked like README.md and tour.md had not been updated.


avi-stripe:
Thanks for this! The tour.md change needs to be made in  https://github.com/stripe/rainier/blob/develop/rainier-docs/src/main/tut/tour.md and then regenerated with `sbt tut`. Could you make that change before we merge?

mfirry:
Oh sure... my bad. Done!

avi-stripe:
Thanks!

Issue #329, opened 2019-02-21
AutoDiff gives erroneous results for functions with conditionals.


arbimo:
It seems that automated differentiation yields erroneous results on expressions with conditionals.
For instance the derivative of `max(x, 0)` should be 0 when `x < 0` and 1 when `x > 0`.
Computing the gradient with automated differentiation results in a function that is always 0.

To reproduce:
```scala
scala> import com.stripe.rainier.compute._
import com.stripe.rainier.compute._

scala> val x = new Variable
x: com.stripe.rainier.compute.Variable = com.stripe.rainier.compute.Variable@49bff24c

scala> val f = x.max(0)
f: com.stripe.rainier.compute.Real = com.stripe.rainier.compute.Lookup@4c405b37

scala> val df = f.gradient.head
df: com.stripe.rainier.compute.Real = com.stripe.rainier.compute.Lookup@1ca97add

scala> val eval = new Evaluator(Map(x -> 1.0))
eval: com.stripe.rainier.compute.Evaluator = com.stripe.rainier.compute.Evaluator@16b2a24b

scala> eval.toDouble(f)
res2: Double = 1.0

scala> eval.toDouble(df)
res3: Double = 0.0 // should be 1.0
```

avi-stripe:
Fixed by #330.

PR #330, opened 2019-02-21
Fix automated differentiation of Lookup and Abs

arbimo:
This PR introduces automated tests for AD by comparing its results to numeric differentiation.
This unveils bugs in the implementation of AD for `abs` and `lookup` which are fixed by this PR (see #329 for the bug report on lookup).

Below are the results of the tests on my machine.
Of the two tests that still fail and for which I would appreciate additional input : 
 -  `pow` seems unrelated to this PR
 - `gamm fit` might fail due to the inaccuracy of numeric differentiation

```
[info] RealTest:
[info] - plus
[info] - exp
[info] - square
[info] - log
[info] - temp
[info] - abs
[info] - max(x, 0)
[info] - max(x, x)
[info] - x > 0 ? x^2 : 1
[info] - x > 0 ? 1 : x + 1
[info] - x > 0 ? x^2 : x + 1
[info] - normal
[info] - logistic
[info] - minimal logistic
[info] - log x^2
[info] - poisson
[info] - 4x^3
[info] - lookup
[info] - exponent sums
[info] - pow *** FAILED ***
[info]   java.lang.ArithmeticException: Cannot take the log of -1.0
[info]   at com.stripe.rainier.compute.RealOps$.unary(RealOps.scala:32)
[info]   at com.stripe.rainier.compute.Evaluator.eval(Evaluator.scala:29)
[info]   at com.stripe.rainier.compute.Evaluator.toDouble(Evaluator.scala:11)
[info]   at com.stripe.rainier.compute.Evaluator.$anonfun$eval$2(Evaluator.scala:26)
[info]   at com.stripe.rainier.compute.Evaluator.$anonfun$eval$2$adapted(Evaluator.scala:26)
[info]   at scala.collection.immutable.List.map(List.scala:290)
[info]   at com.stripe.rainier.compute.Evaluator.eval(Evaluator.scala:26)
[info]   at com.stripe.rainier.compute.Evaluator.toDouble(Evaluator.scala:11)
[info]   at com.stripe.rainier.compute.Evaluator.$anonfun$eval$1(Evaluator.scala:23)
[info]   at com.stripe.rainier.compute.Evaluator.$anonfun$eval$1$adapted(Evaluator.scala:23)
[info]   ...
[info] - gamma fit *** FAILED ***
[info]   8.59681137184154 did not equal 3.523287795242514 [numDiff/diffWithVar, n=1.0] (RealTest.scala:56)
```

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
@@ -103,10 +103,10 @@ private object Gradient {
       case LogOp => gradient.toReal * (Real.one / child.original)
       case ExpOp => gradient.toReal * child
       case AbsOp =>
-        Real.lt(child.original,
+        Real.eq(child.original,
                 Real.zero,
-                gradient.toReal * child.original / child,
-                Real.zero)
+                Real.zero,
+                gradient.toReal * child.original / child)
       case NoOp => gradient.toReal
     }
   }
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
@@ -54,7 +54,7 @@ private object Gradient {
           case l: Lookup =>
             l.table.zipWithIndex.foreach {
               case (x, i) =>
-                diff(x).register(LookupDiff(l, diff(l), i))
+                diff(x).register(LookupDiff(l, diff(l), i + l.low))
                 visit(x)
             }
             visit(l.index)
--- a/rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
+++ b/rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
@@ -4,30 +4,54 @@ import org.scalatest._
 import com.stripe.rainier.core._
 import scala.util.{Try, Success, Failure}
 class RealTest extends FunSuite {
-  def run(description: String)(fn: Real => Real): Unit = {
+  def run(description: String, testDeriv: Double => Boolean = _ => true)(
+      fn: Real => Real): Unit = {
     test(description) {
       val x = new Variable
       val result = fn(x)
+      val deriv = result.gradient.head
+
+      def evalAt(d: Double): Double = Try { fn(Constant(d)) } match {
+        case Success(Infinity)                 => 1.0 / 0.0
+        case Success(NegInfinity)              => -1.0 / 0.0
+        case Success(Constant(bd))             => bd.toDouble
+        case Failure(_: ArithmeticException)   => 0.0 / 0.0
+        case Failure(_: NumberFormatException) => 0.0 / 0.0
+        case x                                 => sys.error("Non-constant value " + x)
+      }
+
       val c = Compiler(200, 100).compile(List(x), result)
+      val dc = Compiler(200, 100).compile(List(x), deriv)
       List(1.0, 0.0, -1.0, 2.0, -2.0, 0.5, -0.5).foreach { n =>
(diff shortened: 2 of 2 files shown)


arbimo:
Last commit uses two-sided numeric differentiation which fixes the error in `gamma fit`.

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
```diff
@@ -54,7 +54,7 @@ private object Gradient {
           case l: Lookup =>
             l.table.zipWithIndex.foreach {
               case (x, i) =>
-                diff(x).register(LookupDiff(l, diff(l), i))
+                diff(x).register(LookupDiff(l, diff(l), i + l.low))
```
avi-stripe, line 57: Yikes. Nice catch!

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
```diff
@@ -102,12 +102,8 @@ private object Gradient {
     def toReal: Real = child.op match {
       case LogOp => gradient.toReal * (Real.one / child.original)
       case ExpOp => gradient.toReal * child
-      case AbsOp =>
-        Real.lt(child.original,
-                Real.zero,
-                gradient.toReal * child.original / child,
-                Real.zero)
-      case NoOp => gradient.toReal
+      case AbsOp => gradient.toReal * child.original / child
```
avi-stripe: Hm, don't we still need to guard against child being zero?

rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
```diff
@@ -4,29 +4,59 @@ import org.scalatest._
...
+          val diffCompiled = dc(Array(n))
+          assertAlmostEqual(diffWithVar,
+                            diffCompiled,
+                            s"[diffWithVar/diffCompiled, n=$n]")
+        }
       }
     }
   }
 
+  def assertAlmostEqual(x: Double,
```
avi-stripe: Is there a reason to keep both this and `assertWithinEpsilon`?

avi-stripe (commented):
Thank you so much for this. Somewhat embarrassing to not have tests on this and have the errors slip through.

avi-stripe:
I'll look into why `pow` is failing

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
arbimo: Yes you're right. I have changed it the last commit. It now does an equality comparison against zero. The original offender was the "lesser than" comparison.

rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
arbimo: Originally, assertWithinEpsilon was not tolerant enough for the inaccuracy of numeric differentiation. Two sided differentiation now provides enough accuracy and I have changed the test to use assertWithinEpsilon everywhere.

avi-stripe:
@arthur-bit-monnot ok, right, so the issue here is that the `pow` test of `x^x` is not continuous for x < 0, so it can't be differentiated. We should either change the function we use to test `pow` or somehow suppress negative values for that one.

If you make that change I'll merge this, or I can do it.

arbimo:
@avi-stripe Ok, I excluded the undefined part of the domain in the tests and everything passes without any workaround now.
I rebased to get three independent commits and it should be ready to be merged now.
Thanks for your help =)

avi-stripe:
Wonderful! Thanks so much for the contribution.

Issue #227, opened 2018-10-15
Add specialized gradients


avibryant:
Stan gets some mileage out of having functions for logp1, log1m, logsumexp, etc, that have specially defined gradients. We should investigate how much performance boost we'd get from doing the same.

PR #331, opened 2019-02-21
Implementation of trigonometric functions

arbimo:
This is a proof of concept PR to gauge the interest in adding common trigonometric functions to `Real`.
It currently only contains a simple implementation for `sin` and `cos` to validate the approach and get early feedback before spending to much time on it.

If there is interest we should:
 - [x] cover all common ones (e.g. see the ones in [spire](https://github.com/non/spire/blob/master/core/src/main/scala/spire/algebra/Trig.scala))
 - [x] add better automated tests for differentiation (e.g. reuse the ones in #330)

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
@@ -107,7 +107,9 @@ private object Gradient {
                 Real.zero,
                 Real.zero,
                 gradient.toReal * child.original / child)
-      case NoOp => gradient.toReal
+      case NoOp  => gradient.toReal
+      case SinOp => gradient.toReal * child.original.cos
+      case CosOp => gradient.toReal * (Real.zero - child.original.sin)
     }
   }
 
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
@@ -110,6 +110,14 @@ private object Gradient {
       case NoOp  => gradient.toReal
       case SinOp => gradient.toReal * child.original.cos
       case CosOp => gradient.toReal * (Real.zero - child.original.sin)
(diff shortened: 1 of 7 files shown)


avi-stripe:
Interesting. Can you say something about the use cases here?

arbimo:
Essentially, I would like to use rainier-core's features of online compilation and automated differentiation for numeric optimization. My current use case is in the control of hybrid systems where trigonometric functions are ubiquitous. Since such trigonometric functions can not be expressed in terms of anything else their absence in Rainier means I cannot use it and would need to maintain a fork.

Overall, implementations of trigo functions should not be intrusive: its mostly additional unary ops as the ones for `sin` and `cos`. With enough automated test the risks would be very low as well.

Another option would be to add an extension point to allow user-defined operations (which could also help for #227) but this would require some non-trivial refactoring and I really believe trigonometric functions have their place in the core.

avi-stripe:
Got it, thanks for the context. I'm +1 to continuing with these. Thanks so much for the contributions!

avi-stripe:
As a side note - I gather from this that you would probably be using the `compute` and `ir` packages but not the `core` package with `RandomVariable` and `Distribution`? If at some point it makes sense to split those out into a separate module I'm sure we can do that.

arbimo:
Great, I'll start working on it.

Indeed, I am only interested in `compute` and `ir` packages. It might make sense to move them to their own module since they are very general purpose and might be useful in a lot of different contexts. However, as long as the rest of  `rainier-core` remains small and with no additional dependencies it is more of a marketing decision with no important technical implications.

arbimo:
@avi-stripe This should be ready for review now.
In the end, there are only intrinsic for (a)sin, (a)cos, (a)tan. Of those only tan could have been defined in term of sin and cos.

I have also added methods on Real for hyperbolic functions that are defined in terms of the existing primitives. Those are not essential to this PR as they only introduce syntactic sugar.

avi-stripe:
Wonderful, thank you!

PR #332, opened 2019-02-23
add exponential, laplace, gamma to sbc tests

avi-stripe:
With #330, these sbc tests are no longer timing out!


mio-stripe:
So good! LGTM

PR #333, opened 2019-02-27
Fixes related to handling unusual double values (+/- infinity and NaN).

arbimo:
Just a few fixes related to edges cases involving infinities and NaN.
One common error was the use of `Constant(d)` to construct a `Real` from a `Double`, this would throw a `NumberFormatException` on NaN and infinities when converting to `BigDecimal`. Using `Real(d)` should handle infinities and throw an arithmetic error on NaN.

### DIFF
--- a/rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
+++ b/rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
@@ -3,6 +3,8 @@ package com.stripe.rainier.compute
 import org.scalatest._
 import com.stripe.rainier.core._
 import scala.util.{Try, Success, Failure}
+import Double.{PositiveInfinity => Inf, NegativeInfinity => NegInf, NaN}
+
 class RealTest extends FunSuite {
   def run(description: String,
           defined: Double => Boolean = _ => true,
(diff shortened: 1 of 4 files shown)


rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
```diff
@@ -129,7 +139,15 @@ class RealTest extends FunSuite {
     Real.gt(x, 0, x * x, x + 1)
   }
 
-  run("normal") { x =>
+  run("normal", defined = !_.isPosInfinity) { x =>
+    // FIXME: letting this be defined at +infinity results in the following error:
+    //     -Infinity did not equal NaN [c/ev, n=Infinity] (RealTest.scala:67)
+    // I.e. two different results from constant folding and evaluation.
+    // same applies to the "normal sum" test case below
+    Normal(x, 1).logDensity(Real(1d))
```
arbimo, line 147: This is a potential problem identified by the additional tests where constant folding and evaluation yield different results.
This is really a edge case but might a symptom of a more general problem.

The problematic case is excluded from the tests since I won't attempt to fix it in this PR but it probably warrants a separate bug report.

avi-stripe:
Thank you!

PR #335, opened 2019-03-05
0.2.2 prep

avibryant:
Some prep for a 0.2.2 release.

* update the list of contributors in the readme to include @arthur-bit-monnot 
* depend on a published shaded-asm artifact rather than on the sbt sub-module. It's a little wonky here the way we still publish a new asm artifact for every rainier release even though it never changes, and that we just depend on a random previously published version, but this was the quickest way to improve things and we can always make it better later.
* uncomment `tutTargetDirectory` because our doc gen should be deterministic now. @andyscott, is this actually checked by CI or is there something else I'd need to enable there?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -134,8 +134,7 @@ lazy val rainierDocs = project.
     scalacOptions in Tut ~= {
       _.filterNot(sc => sc.contains("-Ywarn-unused") || sc == "-Yno-predef" )
     },
-    // todo: uncomment once docs generation is deterministic
-    // tutTargetDirectory := (baseDirectory in LocalRootProject).value / "docs"
+    tutTargetDirectory := (baseDirectory in LocalRootProject).value / "docs"
   ).
   settings(unpublished)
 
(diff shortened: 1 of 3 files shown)


andyscott:
Uncommenting that line is all you should need to do.

build.sbt
```diff
@@ -73,23 +73,20 @@ lazy val V = new {
...
   settings(
-    bazelRuleDeps := (Deps(Compile) - ScalaLib(Compile)) +
-      BazelDep("//.rainier-shaded-asm", "asmTreeShaded") +
-      BazelDep("//.rainier-shaded-asm", "asmShaded") -
-      BazelDep("//.rainier-shaded-asm", "shadedAsm"),
     libraryDependencies ++= Seq(
       "com.google.flogger" % "flogger" % V.flogger,
-      "com.google.flogger" % "flogger-system-backend" % V.flogger)
+      "com.google.flogger" % "flogger-system-backend" % V.flogger,
+      "com.stripe" % "rainier-shaded-asm_6.0" % V.shadedAsm)
```
andyscott, line 89: 👍 nice, this will make things easier

andyscott (approved):

PR #336, opened 2019-03-13
fixed distribution, added tests

mikeheaton-stripe:
Fixes a bug with the normal approximation on the binomial distribution. We were scaling by the variance instead of scaling by the standard deviation, leading to much wider distributions than are proper.

Added a test to `DiscreteTest.scala` to test the mean and variance of distributions against known functions.

r? @mio-stripe @avi-stripe


rainier-tests/src/test/scala/com/stripe/rainier/core/DiscreteTest.scala
```diff
@@ -0,0 +1,188 @@
+package com.stripe.rainier.core
```
mikeheaton-stripe, line 1: This test file has been removed from the Develop branch, not sure why. I can take out the `check` test and leave just the test I wrote, if you want.

avibryant: It may have been taken out because it's assumed that `SBCTest` replaces it. Can you check if https://github.com/stripe/rainier/blob/develop/rainier-tests/src/test/scala/com/stripe/rainier/core/SBCTest.scala#L23 (which has been commented out) works now?

avibryant:
Nice catch!

mikeheaton-stripe:
After
image
Before
image

rainier-tests/src/test/scala/com/stripe/rainier/core/DiscreteTest.scala
mikeheaton-stripe: It doesn't, but if it was being generated off of an incorrect goldset, maybe we need to update the goldset? (I'm not familiar with these sbc tests!)

avibryant: yeah, you need to `run` instead of `test`. Right now it's showing overdispersion (which is consistent with what you found). Hopefully when you run it, it looks more uniform.

Screen Shot 2019-03-12 at 5 50 01 PM

avibryant: sorry, my screenshot was over-cropped, but note that you have to be in `project rainierTests`

mikeheaton-stripe: Uhh, 
image
the bottom line overflowed and made the others look odd. Basically the bottom line hugely overflowed, how unfortunate.

avibryant: Ok, so it still doesn't "pass" that test...

avibryant: Specifically it's showing that when you generate samples from the Binomial with a true value of `x`, and then try to infer `x'` from those samples, your inferred `x'` is biased to be lower than `x`.

avibryant: If we presume that the `logDensity` and hence the inference is correct, that would imply that our generation is now biased to be too low.

mikeheaton-stripe: Strange that it's now a one-sided error

avibryant: @mikeheaton-stripe should we discuss in https://gitter.im/com_stripe_rainier/Lobby ?

mikeheaton-stripe: :blue-portal:

avibryant:
I would really love to understand what's happening with SBC here, but this is still a major improvement and once CI is green we should merge.

PR #338, opened 2019-04-05
warn on methods getting too large

avibryant:
This uses some janky reflection to reach into the internals of ASM6 and pull out the bytecode sizes of individual methods from `ClassWriter`. This lets us produce a warning if the sizes start to exceed the default max JIT size.

Obviously something using a public interface would be better, and this will need to be updated if we ever upgrade the ASM version, but it's a tiny and encapsulated piece of code that is very valuable as a debugging tool, so I think it's worth it.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/internal/asm/MethodSizer.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/internal/asm/MethodSizer.scala
@@ -1,0 +1,16 @@
+package com.stripe.rainier.internal.asm
+
+object MethodSizer {
+  val firstMethodField = classOf[ClassWriter].getDeclaredField("firstMethod")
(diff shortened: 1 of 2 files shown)


mio-stripe:
LGTM!

gkossakowski:
this hack is fine but why not have (implement) another visitor sitting along the CodeWriter that will compute just method sizes?

gkossakowski:
it actually exists already: https://asm.ow2.io/javadoc/org/objectweb/asm/commons/CodeSizeEvaluator.html

i would vote for using it instead of reflection hacks

avibryant:
@gkossakowski oh, nice find!
We should definitely switch to that. I think I'll leave it as an easy contribution for someone else, for a while anyway, and circle back to it later if nobody else takes.

PR #337, opened 2019-03-29
Update bintray server_urls to fetch via TLS

nati-stripe:
Manually updating the WORKSPACE file to ensure all the server urls are accessing over TLS.

PR #339, opened 2019-04-12
Update resolvers to fetch dependencies via TLS

mio-stripe:
Do-over of https://github.com/stripe/rainier/pull/337 following @beala-stripe 's suggestion https://github.com/stripe/rainier/pull/337#issuecomment-478730807 to update the resolvers rather than editing the WORKSPACE file by hand.


mio-stripe:
@beala-stripe mind giving me some pointers in rerunning sbt-bazel so I can make sure this works?

beala-stripe:
`bazelGenerate` is the sbt task that generates the bazel files.

You can:
```
cd rainier
sbt bazelGenerate
```

Or run the sbt repl and run `bazelGenerate` inside it.

mio-stripe:
Sweet, thanks!
r? @avi-stripe @beala-stripe

avibryant:
LGTM

rainier-core/BUILD
```diff
@@ -8,15 +8,12 @@ load(
 scala_library(
   name = 'rainierCore',
   deps = [
-    '//.rainier-shaded-asm:asmShaded',
-    '//.rainier-shaded-asm:asmTreeShaded',
+    '//external:jar/com/google/code/findbugs/jsr305',
     '//external:jar/com/google/flogger/flogger',
-    
-  ],
-  runtime_deps = [
```
beala-stripe: It looks like these runtim_deps got dropped. Is it possible this file was edited by hand before?

mio-stripe: whoa, good catch. Dunno, but I'll just add them back.

mio-stripe: Running `sbt bazelGenerate` on current develop, I get https://github.com/stripe/rainier/compare/mio/no-https which also deletes those deps. I know shaded-asm has some complicated history, did we intentionally change this? (cc: @avi-stripe)

nati-stripe:
Thanks for tackling this the correct way @mio-stripe @beala-stripe !

PR #342, opened 2019-06-28
some utilities for interactive use

avibryant:
This adds some summarization methods to the `repl` package (`precis`, `hdpi`), and some plots to the `Jupyter` object (`scatter`, `scatterLines`, `histogram`, `contour`), that are based on the exploratory code in "Statistical Rethinking".


mio-stripe:
Nice!
LGTM

PR #344, opened 2019-07-04
scatterShaded

avibryant:
This adds a `scatterShaded` method to the `Jupyter` tools in the style of Statistical Rethinking. It takes the same interface as `scatterLines`, where you provide both raw 2D data and a function to do posterior predictions from x => y, but rather then drawing a line for each sample, it shades the HDPI region of those samples, and then draws a single line through the center of that region.

I originally attempted to do a full density heatmap but that was taking too much time to do correctly, so I settled for this.

Screenshot from 2019-07-04 13-54-12

### DIFF
--- a/rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
+++ b/rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
@@ -6,7 +6,10 @@ import com.cibo.evilplot.plot._
 import com.cibo.evilplot.numeric._
 import com.cibo.evilplot.colors._
 import com.cibo.evilplot.plot.renderers._
+import com.cibo.evilplot.plot.renderers.SurfaceRenderer.SurfaceRenderContext
 import com.cibo.evilplot.plot.aesthetics.DefaultTheme._
+import com.cibo.evilplot.plot.aesthetics.Theme
+import com.stripe.rainier.repl.hdpi
 
 object Jupyter {
   implicit val extent = Extent(400, 400)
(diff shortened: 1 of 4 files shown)


avibryant:
I also snuck in a evilplot version bump and some histogram tweaks.

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -143,4 +187,39 @@ object Jupyter {
     baos.close
     array
   }
+
+  //copied from private evilplot class
+  private case class SurfacePlotRenderer(
```
andrew-stripe: I don't see this referenced anywhere. Was it part of an older iteration on this code?

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -88,6 +88,50 @@ object Jupyter {
...
+    }
+  }
+
+  def scatterShaded(seq: Seq[Map[String, Double]], xKey: String, yKey: String)(
+    fn: Double => Seq[Double])(implicit oh: OutputHandler,
+                               extent: Extent): Unit = {
+    val xy = seq.map { s => (s(xKey), s(yKey)) }
+    val scatter = scatterPlot(xy)  
+  
+    val minX = BigDecimal(scatter.xbounds.min)
```
andrew-stripe: Might make sense to move this intervals calculation into `shadedRenderer` since you have the bounds you need in there. I think then `shadedRenderer` would take a function `Double => Seq[Double, (Double, Double)]`

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -88,6 +88,50 @@ object Jupyter {
...
+    val maxX = BigDecimal(scatter.xbounds.max)
+    val grid = (maxX - minX) / 20.0
```
andrew-stripe: Hardcoded `20` seems like it'll bite us working with curves, but I suppose can address that when the time comes.

andrew-stripe (approved):
Some minor comments, no need to re-r if they make sense. 

I don't love that these methods don't compose like the original R/EvilPlot ones such that every variation on a plot needs its own method. It feels like you'll eventually want to do something more like:
```
val scatter = scatterPlot(...)
val line = linePlot(...)
val shading = shadedPlot(...)
show(List(scatter, line, shading), xKey, yKey)
```

Not sure about the compromises there between how much EvilPlot to expose vs how much to tailor things to the plots the book uses.

avibryant:
@andrew-stripe yeah, I've been feeling the same thing about composability. There are some tradeoffs and design issues we should talk through but I think it's better to defer merging this until we work it out.

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -6,125 +6,146 @@ import com.cibo.evilplot.plot._
...
-        .yLabel("Frequency"))
-  }
+  def density[N](seq: Seq[N],
+                 xLabel: String = "",
+                 bounds: Option[Bounds] = None)(implicit num: Numeric[N],
+                                                theme: Theme): Plot =
+    Histogram(seq.map { n =>
+      num.toDouble(n)
+    }, binningFunction = Histogram.density, xbounds = bounds)
+      .xLabel(xLabel)
```
andrew-stripe: This shouldn't set xlabel/yabel/xaxis/yaxis/frame anymore right?

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -6,125 +6,146 @@ import com.cibo.evilplot.plot._
...
-    show(
-      Overlay
-        .fromSeq(scatter :: functionPlots)
-        .xAxis()
-        .yAxis()
-        .frame()
-        .xLabel(xKey)
-        .yLabel(yKey))
+
+  def lines(xbounds: Bounds)(fn: Double => Seq[Double])(
```
andrew-stripe, line 72: Not sure it's important to have a dedicated `lines` rather than generating N separate plots with `line`, but doesn't hurt so I don't have a strong opinion.

andrew-stripe:
Few minor comments but the new API looks good!

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
avi-stripe: Yeah, fixed (and added a `show` variant that only does an x-axis to get around the need to write "Density" in every time)

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
avi-stripe: I think it's convenient enough to be worth leaving

PR #346, opened 2019-07-15
refactor to allow Predictor.fit

avi-stripe:
* narrow the interface of `Predictor` to allow non-`Encoder` implementations
* provide a scalar version of `Likelihood.fit`
* use these to provide an immediate `Predictor.fit` that does not use any vector/placeholder logic, for small datasets


avi-stripe:
cc @mio-stripe

andrew-stripe (approved):
Interface looks reasonable but I'm not familiar enough with Rainier internals to review the implementation deeply.

PR #347, opened 2019-07-16
fix Predictor nesting

avi-stripe:


avi-stripe:
Changing the title of this PR since I accidentally pushed something else to it :) This makes sure there's a `ToLikelihood` for `Predictor` which allows them to be nested properly. It also random and incidentally adds a color parameter to `scatter`.

andrew-stripe (approved):
Approved with the caveat that I don't understand the ToLikelihood bits in detail

PR #348, opened 2019-07-18
bail out of HMC if we get to 0 step size

avi-stripe:
This stops trying to find a reasonable step size if our stepSize gets down to 0, since we're obviously never going to get anywhere useful if we're not moving. This will cause a WARNING to be logged further down the call stack.


andrew-stripe (approved):
I don't see what causes the warning to get logged but assuming you've tested it this looks good.

I assume there are certain cases (all cases?) where a step-size of zero means the result won't converge to something meaningful. Should that error be expressed more clearly as an exception or something?

avibryant:
Yes, in all cases the results here will not be useful or meaningful. As a matter of scala style I'd rather not raise an exception for this. I'm not totally sure what to suggest; maybe we can introduce a `SampleResult` ADT that has success and failure variants, but do we still want a simple method where you don't have to check that?

(Either way that should be a follow up PR).

PR #350, opened 2019-07-31
sbt bazelGenerate

metcalf:
r? @avibryant


rainier-core/BUILD
```diff
@@ -8,14 +8,12 @@ load(
 scala_library(
   name = 'rainierCore',
   deps = [
-    '//.rainier-shaded-asm:asmShaded',
```
metcalf, line 11: The changes to `-core` seem most notable to me. Does this look right?

avi-stripe: It looks like this is because we switched to treating the shaded jar as an "external" dep. If `bazel build ...` works, it's fine.

rainier-scalacheck/BUILD
```diff
@@ -10,6 +10,8 @@ scala_library(
   deps = [
     '//external:jar/com/google/code/findbugs/jsr305',
     '//external:jar/com/google/flogger/flogger',
+    '//external:jar/com/google/flogger/flogger_system_backend',
```
avi-stripe, line 13: Worried about this as a dep rather than a runtime dep. Might be worth moving manually.

PR #353, opened 2019-08-01
Remove Bazel build

metcalf:
This removes the Bazel build generated by sbt-bazel in favor of just publishing packages for internal consumption. I confirmed that I can build Zoolander against this.

cc @mio-stripe


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.2.3-SNAPSHOT"
```
metcalf, line 1: Do we want to retain the `SNAPSHOT` suffix on the `develop` branch? Publish this as rc1 and then immediately add `-SNAPSHOT`?

avi-stripe: yes, I think we can publish `0.2.3-rcX` and then go back to `0.2.3-SNAPSHOT`.

build.sbt
```diff
@@ -221,39 +188,7 @@ lazy val shadedAsm = project.
...
-        |jar_jar(
-        |    name = "asmTreeShaded",
-        |    input_jar = "@org_ow2_asm_asm_tree//jar",
-        |    rules = "shade_rule",
-        |    visibility = [
-        |      '//visibility:public',
-        |    ],
-        |)
-      """.stripMargin)
+    releaseVersion := { ver => ver }
```
avi-stripe, line 191: I don't know sbt well enough but could this be used to avoid us having to hand-modify the `version` all the time?

avi-stripe:
I think it's ok to merge this as is and followup with the version numbering stuff

avi-stripe (approved):

metcalf:
I think there's value in committing the actual RC version changes to the develop branch to make it easier to figure out what a particular artifact came from. I'd be inclined to update the version to `0.2.3-rc1-SNAPSHOT` and publishing this internally to that end.

PR #354, opened 2019-08-09
bare bones RVG

avi-stripe:
This is the simplest possible implementation of an `RVG[T]` type, which is a monad-transformer-style stacking of `RandomVariable[Generator[T]]`. There are lots of convenience methods missing, eg, for converting from `RandomVariable[T:ToGenerator]`, or for sampling `RVG[T]`.

This implementation is technically a functor rather than a monad, because it does not provide `flatMap`. The reasoning is that it doesn't make any sense to create a new `RandomVariable` sequentially after a `Generator` (because `RandomVariable` is about creating/updating the posterior, and `Generator` is about predicting *from* the posterior). However, it does provide a `flatMapGen` for conveniently flatmapping the internal Generator from the stacked object.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/RVG.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/RVG.scala
@@ -1,0 +1,21 @@
+package com.stripe.rainier.core
+
+case class RVG[T](value: RandomVariable[Generator[T]]) {
+  def map[U](fn: T => U): RVG[U] = RVG(value.map(_.map(fn)))
+
+  def zip[U](other: RVG[U]): RVG[(T, U)] =
+    RVG(value.zip(other.value).map { case (tg, ug) => tg.zip(ug) })
+
+  def flatMapGen[U](fn: T => Generator[U]): RVG[U] =
+    RVG(value.map(_.flatMap(fn)))
+}
+
+object RVG {
+  def traverse[T](seq: Seq[RVG[T]]): RVG[Seq[T]] =
+    RVG(
+      RandomVariable
+        .traverse(seq.map(_.value))
+        .map { gs =>
+          Generator.traverse(gs)
+        })
+}


rainier-core/src/main/scala/com/stripe/rainier/core/RVG.scala
```diff
@@ -0,0 +1,21 @@
...
+  def flatMapGen[U](fn: T => Generator[U]): RVG[U] =
+    RVG(value.map(_.flatMap(fn)))
+}
+
+object RVG {
+  def traverse[T](seq: Seq[RVG[T]]): RVG[Seq[T]] =
+    RVG(
+      RandomVariable
+        .traverse(seq.map(_.value))
+        .map { gs =>
```
andrew-stripe, line 18: I noticed in other parts of the codebase as well that you tend not to write these cases as `.map(Generator.traverse(_))`. Any particular reason?

rainier-core/src/main/scala/com/stripe/rainier/core/RVG.scala
```diff
@@ -0,0 +1,21 @@
+package com.stripe.rainier.core
+
+case class RVG[T](value: RandomVariable[Generator[T]]) {
```
andrew-stripe, line 3: Probably worth a quick comment explaining what this is.

andrew-stripe (approved):
This looks helpful, thanks! My only concern is whether RVG is too opaque a name. RandomVariableGenerator feels weird since it's not a generator of random variables. RandomVariableT or RandomVariableTransformer captures the fact that this is a transformer (of which it just so happens to only make sense to implement for generators).

rainier-core/src/main/scala/com/stripe/rainier/core/RVG.scala
avi-stripe: I'm gunshy of the rules about `_` inside nested scopes so I try to avoid using them except for something like `.map(_.foo)`.

rainier-core/src/main/scala/com/stripe/rainier/core/RVG.scala
avi-stripe: Agreed but will leave for a follow up PR.

avi-stripe:
Properly it should be `GeneratorT` which means that `RandomVariableT` is too confusing; also that implies a generality we don't support. I think this is a case where we should just embrace the slight opacity; `RVG` is easy to type and say and makes sense once you know what it is.

PR #355, opened 2019-08-09
bare bones Data

avi-stripe:
This is a bare bones `Data` class which allows for `Data(seq).update(dist)` instead of `dist.fit(seq)`. It also allows `Data(pairs).update{k => distFor(k)}` instead of `Predictor.fit(pairs){k => distFor(k)}`.

The intent here is to make introducing observations more consistent and flexible instead of the current `dist.fit`. `predictor.fit`, and `Predictor.fit` variations. To actually replace `Predictor` we'll need to introduce a `Fn` type (PR to follow separately).

As with the `RVG` PR, there's a lot of convenience methods we should add here, just trying to get the basics settled.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Data.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Data.scala
@@ -1,0 +1,14 @@
+package com.stripe.rainier.core
+
+case class Data[T](seq: Seq[T]) {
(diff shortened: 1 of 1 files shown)


andrew-stripe (commented):
The caller/callee reads a bit backwards to me here. I read `data.update(fn)` as "update data using the arguments passed". Putting the update methods on Fn would avoid adding another noun to the API. Were there other reasons you wanted this Data type to hang methods off of?

avi-stripe:
Closing in favor of standardizing on `RandomVariable.fit`

PR #357, opened 2019-08-12
Add support for traversing maps in RandomVariable and Generator

metcalf:
This adds support for Map types to `RandomVariable.traverse` and `Generator.traverse`. For both, we assume that the keys will be constants and the values will be RandomVariable/Generator. Additionally, I've changed the signature to `Generator.traverse` to take an implicit `ToGenerator` for elements of the sequence rather than requiring it to already be a sequence of generators.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
@@ -104,31 +104,26 @@ object Generator {
 
   def traverse[T, U](seq: Seq[T])(
       implicit toGen: ToGenerator[T, U]
-  ): Generator[Seq[U]] =
+  ): Generator[Seq[U]] = {
+    val asGen = seq.map(toGen(_))
     new Generator[Seq[U]] {
-      val requirements: Set[Real] = seq.flatMap(toGen(_).requirements).toSet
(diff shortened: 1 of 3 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -102,12 +102,31 @@ object Generator {
...
 
-  def traverse[T](seq: Seq[Generator[T]]): Generator[Seq[T]] =
-    new Generator[Seq[T]] {
-      val requirements: Set[Real] = seq.flatMap(_.requirements).toSet
-      def get(implicit r: RNG, n: Numeric[Real]): Seq[T] =
+  def traverse[T, U](seq: Seq[T])(
+      implicit toGen: ToGenerator[T, U]
+  ): Generator[Seq[U]] =
+    new Generator[Seq[U]] {
+      val requirements: Set[Real] = seq.flatMap(toGen(_).requirements).toSet
```
avi-stripe: seems like you could do the `toGen` conversion once and stash that rather than doing it both here and every time there's a `get`?

avi-stripe (changes requested):
One small comment, fine to merge once addressed.

PR #359, opened 2019-08-21
whiskers plot

avi-stripe:
This adds an option to use evilplot's default BoxPlot to draw box-and-whiskers diagrams from the same sample format as we use in `precis`. Here's an example from the book.

Screen Shot 2019-08-21 at 1 12 49 PM

### DIFF
--- a/rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
+++ b/rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
@@ -8,6 +8,8 @@ import com.cibo.evilplot.colors._
 import com.cibo.evilplot.plot.renderers._
 import com.cibo.evilplot.plot.aesthetics._
 
+import com.stripe.rainier.repl.hdpi
+
 object Jupyter {
   val font =
     java.awt.GraphicsEnvironment.getLocalGraphicsEnvironment.getAvailableFontFamilyNames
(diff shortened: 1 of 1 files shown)


rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -91,6 +91,16 @@ object Jupyter {
                             Some(xbounds))
     })
 
+  def whiskers[K, N](samples: Seq[Map[String, Double]]): Plot = {
```
andrew-stripe: It looks like the type parameters here aren't used. Should they be in place of String and Double?

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -91,6 +91,16 @@ object Jupyter {
...
+    val labels = samples.head.keys.toList.sorted
+    val data = labels.map { k =>
+      samples.map { m =>
+        m(k)
+      }
+    }
+    BoxPlot(data).standard(xLabels = labels)
+  }
```
andrew-stripe, line 186: I'd suggest implementing this form as well and having the current form call this:
```
def whiskers(data: Map[String, Seq[Double]]): Plot =
  BoxPlot(data.values).standard(xLabels = data.keys)
```

andrew-stripe:
How well does this scale visually to plots with a lot of labels like figure 5.6(b)?
image

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
avibryant: that was from an earlier version and I forgot to clean them up. The types in this currently exactly match `precis`, which seems appropriate. We may in a later PR want to make both of those methods more flexible.

avibryant:
@andrew-stripe the visual scaling question is a good one. Rotating the whole thing 90 degrees, as the book does, would definitely help with that; it's not obvious how to do that with evilplot but I can work on it. Again, I'd suggest doing that as a later PR to this one.

andrew-stripe:
With the horizontal layout, do the box widths scale down as the amount of data increases? That might be enough for most of the charts we need. I think that's fine to punt to a future PR. I'd still suggest doing https://github.com/stripe/rainier/pull/359/files#r316399590 here.

avibryant:
Ok, now looks like this.
image

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -94,11 +96,90 @@ object Jupyter {
   def whiskers[K, N](samples: Seq[Map[String, Double]]): Plot = {
```
andrew-stripe: This should use `Seq[Map[K, N]]` to match the other signature right?

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -91,6 +93,95 @@ object Jupyter {
...
+  }
+
+  def whiskers[K, N](seq: Seq[(K, Seq[N])], ybounds: Option[Bounds])(
+      implicit num: Numeric[N],
+      theme: Theme): Plot = {
+    val data = seq.map { case (_, ns)  => ns.map(num.toDouble) }
+    val labels = seq.map { case (k, _) => k.toString }
+
+    val xb = Bounds(0, seq.size.toDouble)
+    val yb = ybounds.getOrElse(
```
andrew-stripe: These y bounds include the full range of data which we're not currently plotting anything for. Maybe this should use the BoxPlotSummaryStatistics to bound on the highest upperWhisker and lowest lowerWhisker?

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -91,6 +93,95 @@ object Jupyter {
...
+      Bounds(
+        data.flatten.reduceOption[Double](math.min).getOrElse(0),
+        data.flatten.reduceOption[Double](math.max).getOrElse(0)
+      ))
+
+    val boxContexts = data.zipWithIndex.map {
+      case (dist, index) =>
+        if (dist.nonEmpty) {
+          val summary = BoxPlotSummaryStatistics(dist, (0.25, 0.50, 0.75))
```
andrew-stripe: Should plausibly make these quantiles an optional argument as with https://github.com/cibotech/evilplot/blob/a44bdd84ceca4e198ff5527cf1aaed017febd91e/shared/src/main/scala/com/cibo/evilplot/numeric/BoxPlotSummaryStatistics.scala#L63

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -91,6 +93,95 @@ object Jupyter {
...
+
+    Plot(
+      xb,
+      yb,
+      BoxPlotRenderer(
+        boxContexts,
+        BoxRenderer.custom { (extent, ctx) =>
+          val stats = ctx.summaryStatistics
+          val mu = stats.allPoints.sum / stats.allPoints.size
+          val (low, high) = hdpi(stats.allPoints)
```
andrew-stripe: I think you want stats.{lowerQuantile,upperQuantile} here instead of hdpi. If we want different quantiles we should probably make that a parameter.

andrew-stripe (approved):
Yowzas this escalated fast! Some material comments but I think it's fine to land without a re-review after addressing those since I'm going to be out until Tuesday.

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
avi-stripe: see previous comment about precis

rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
avi-stripe: that's fair. I do want to make the default quantiles match the book's HDPI range. But it doesn't have to be the hdpi algo.

avi-stripe:
I ended up going for maximum control and having the non-convenience method just take `BoxPlotSummaryStatistics` objects directly, which can be prepared however you like. The `precis`-alike convenience method uses `hdpi`.

PR #361, opened 2019-08-27
Implement generator traversal using implicits

metcalf:
This re-implements `Generator.traverse` based on a `GeneratorTraversal` implicit trait. This allows for defining traversals for custom classes as demonstrated in the test. I'm pretty sure you could use shapeless with this to implement traversal for any case class where each member has a `ToGenerator`.

Assuming this looks good, I'll implement the same thing for RandomVariable.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
@@ -126,9 +126,9 @@ final case class Multinomial[T](pmf: Map[T, Real], k: Real)
 }
 
 object Multinomial {
-  def optional[T](pmf: Map[T,Real], k: Real): Multinomial[Option[T]] = {
+  def optional[T](pmf: Map[T, Real], k: Real): Multinomial[Option[T]] = {
     val total = Real.sum(pmf.values)
-    val newPMF = pmf.map{case (t,p) => Option(t) -> p} + (None -> (Real.one - total))
+    val newPMF = pmf.map { case (t, p) => Option(t) -> p } + (None -> (Real.one - total))
     Multinomial(newPMF, k)
   }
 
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -126,9 +126,9 @@ final case class Multinomial[T](pmf: Map[T, Real], k: Real)
 }
 
 object Multinomial {
-  def optional[T](pmf: Map[T,Real], k: Real): Multinomial[Option[T]] = {
```
metcalf, line 129: Unrelated but running the code picked up a linter change from the previous PR.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -102,28 +102,43 @@ object Generator {
...
-    val asGen = seq.map(toGen(_))
-    new Generator[Seq[U]] {
-      val requirements: Set[Real] = asGen.flatMap(_.requirements).toSet
-      def get(implicit r: RNG, n: Numeric[Real]): Seq[U] = asGen.map(_.get)
+  def traverse[T, U](t: T)(
+      implicit traversal: GeneratorTraversal[T, U]
+  ): Generator[U] = traversal(t)
+}
+
+trait GeneratorTraversal[-T, U] {
```
metcalf, line 110: Should this be a verb instead of a noun? Better wording?

avi-stripe:
I'd love to get @non or @andyscott's take on this.

avi-stripe:
What I'm not sure about here is whether it's actually desirable to have a generic `Generator.traverse` which can handle arbitrary data structures, or whether it's better to have a small number of specific variants like `traverseMap`.

andyscott:
I would flip this around and use existing machinery in cats:

```
traverse

avibryant:
Closing this since I'm pretty sure we don't think this is the right approach.

PR #362, opened 2019-08-28
Make Encoder support arbitrary Numerics

metcalf:
I think this is both simpler and more general. This was motivated by noticing there's no `Encoder[Long]` when I was working on making `Discrete` a `Distribution[Long]`.


avi-stripe:
Ah yes this is great. LGTM.

PR #363, opened 2019-08-28
Switch Discrete to be a Distribution[Long]

metcalf:
We've run into cases where we may want to output large values from `Discrete` so this switches it to use `Long`. This is a breaking change so it'll merge into the 0.3-dev branch.


avi-stripe:
LGTM once it's green.

andrew-stripe:
@avi-stripe I assume it's OK to commit the results of running `sbt tut`?

avi-stripe:
@andrew-stripe yep.

PR #364, opened 2019-08-29
L-BFGS optimization

avi-stripe:
This is a WIP, but I'm also open to the idea that we should just merge it now rather than sinking more time into it.

This includes:

- an `optimize` method on `RandomVariable` which returns the MAP instead of a full posterior sample
- an `Optimizer` class in a new package which implements this, as a thin wrapper around LBFGS
- the `LBFGSOrig.java` which is a fairly straight extraction from https://github.com/dbtsai/lbfgs which in turn is extracted from RISO which in turn is a mechanical translation of the canonical FORTRAN implementation. This basically uses global (ie static) variables for all state. We keep this around, temporarily, to verify that the refactoring in `LBFGS` has not changed the logic at all.
- `LBFGS` which is still pretty awful but is at least thread-safe and does what it can to make the code a little more sane and readable. There's a lot more work that could be done here if we think it's worth it.


rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -82,6 +83,11 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       }
   }
 
+  def optimize[V](implicit rng: RNG, tg: ToGenerator[T, V]): V = {
+    val array = Optimizer.lbfgs(density)
+    tg(value).prepare(targetGroup.variables).apply(array)
```
avi-stripe, line 88: As a future refinement, I suspect we should introduce a `Decoder[T,V]` and require that instead of `ToGenerator[T,V]` here, where `Decoder` does *not* take a `RNG`. It's possible this method will still need a `RNG` if we want random init of the param vector before optimization.

rainier-core/src/main/scala/com/stripe/rainier/optimizer/LBFGSOrig.java
```diff
@@ -0,0 +1,1030 @@
+/* Extracted and adapted from:
```
andrew-stripe: Rather than merging this into the source, can/should we maintain a copy of this in the tests and write some explicit tests comparing their basic behavior?

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -82,6 +83,11 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       }
   }
 
+  def optimize[V](implicit rng: RNG, tg: ToGenerator[T, V]): V = {
```
andrew-stripe, line 86: Maybe this should take the optimizer as a parameter that defaults to `Optimizer.lbfgs`? That would allow us to more easily swap it out in tests.

rainier-core/src/main/scala/com/stripe/rainier/optimizer/LBFGS.java
```diff
@@ -0,0 +1,755 @@
+/* Heavily modified from:
+ * RISO: an implementation of distributed belief networks. (Copyright 1999, Robert Dodier).
+ *
+ * License: Apache license, version 2.
+ */
+package com.stripe.rainier.optimizer;
+
+class LBFGS
```
andrew-stripe, line 8: Per IRL, this should be private so no one can depend on the implementation and probably deserves a comment about being a pretty minimal port just to  get thread safety.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
avi-stripe: I ended up writing the tests at a slightly lower level anyway. We may well end up wanting to support multiple optimizers later but I think it's premature for now.

rainier-core/src/main/scala/com/stripe/rainier/optimizer/LBFGS.java
avi-stripe: Java's default is package-private, which is what we want.

avi-stripe:
@andrew-stripe I think this is ready to merge, if you want to re-review

rainier-core/src/main/scala/com/stripe/rainier/optimizer/Optimizer.scala
```diff
@@ -0,0 +1,25 @@
+package com.stripe.rainier.optimizer
+
+import com.stripe.rainier.sampler.DensityFunction
+
+object Optimizer {
+  def lbfgs(df: DensityFunction): Array[Double] = {
```
andrew-stripe, line 6: Given how opaque the LBFGS library is, it might be worth a lot of comments in this method explaining how it's using the library.

andrew-stripe (approved):

rainier-core/src/main/scala/com/stripe/rainier/optimizer/Optimizer.scala
avi-stripe: Will take as a follow-up.

PR #365, opened 2019-08-29
Events.observe

avi-stripe:
This standardizes the ways to observe event data in a single `Events` object with a set of `observe` methods. The implementations here are not new (just moved from other places); the thing to focus on is the naming and choice of signatures. Once choice this makes is for the return value to be to what I believe is the most useful posterior-predictive generator object (eg, for the function variants, it's from `Seq[X] => Generator[Seq[(X,Y)]]`, which (given new feature vectors) produces generated predictive output in the same form as the observed data.)

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
@@ -1,0 +1,32 @@
+package com.stripe.rainier.core
+
+object Events {
+    def observe[T](seq: Seq[T], dist: Distribution[T]): RandomVariable[Generator[T]] =
+        dist.fit(seq).map(_.generator)
+
+    def observe[X,Y](seq: Seq[(X,Y)])(fn: X => Distribution[Y]): RandomVariable[Seq[X] => Generator[Seq[(X,Y)]]] = {
+        val rvs = seq.map { case (x, z) => fn(x).fit(z)}
+        RandomVariable.traverse(rvs).map { _ =>
+            (seq: Seq[X]) => 
+                Generator.traverse(seq.map{x =>
+                    fn(x).generator.map{y => x -> y}
+                })
+      }
+    }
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
```diff
@@ -0,0 +1,32 @@
...
+        val lh = Fn.toLikelihood[X,Distribution[Y],Y](implicitly[ToLikelihood[Distribution[Y],Y]])(fn)
+        lh.fit(seq).map{_ =>
+            (seq: Seq[X]) => 
+                Generator.traverse(seq.zip(fn(seq)).map{case (x,dist) =>
+                    dist.generator.map{y => x -> y}
+                })       
+        }
+    }
+
+    def observe[X,Y](xs: Seq[X], ys: Seq[Y])(fn: X => Distribution[Y]): RandomVariable[Seq[X] => Generator[Seq[(X,Y)]]] =
```
andrew-stripe: Is this a common enough case that's it's worth providing dedicated sugar vs. just expecting the caller to `zip`? I might be inclined to leave this off and only include it if we find an annoyingly high frequency of zipping.

rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
```diff
@@ -0,0 +1,32 @@
+package com.stripe.rainier.core
+
+object Events {
+    def observe[T](seq: Seq[T], dist: Distribution[T]): RandomVariable[Generator[T]] =
```
andrew-stripe: IRL we discussed having these be something like `def observe[T, D <: Distribution[T]](seq: Seq[T], dist: D): RandomVariable[D]`

avi-stripe:
merging after IRL convo with @andrew-stripe

PR #366, opened 2019-08-30
Make Generator's Monad instance stack safe

sritchie:
This PR changes the implementation of `Generator[T]` to a sealed trait with two concrete instances - `Constant[T]` and `From[T]`. This allowed me to rewrite the `tailRecM` method of the monad instance to make it stack safe.

Thanks to cats, adjusting the tests was super easy! I just changed `.stackUnsafeMonad` to `.monad` (after verifying that it failed before my changes) and, 💥 , all was well.

I rewrote much of `Generator.scala` to use `Concrete[T]` and `From[T]` instead of creating anonymous instances. I'll comment below on the changes.

### DIFF
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -1,10 +1,9 @@
 package com.stripe.rainier.cats
 
+import com.stripe.rainier.compute.Real
 import com.stripe.rainier.core.Generator
 import com.stripe.rainier.core.RandomVariable
-import com.stripe.rainier.compute.Real
 import com.stripe.rainier.sampler.RNG
-
 import _root_.cats.Monad
 
 import scala.annotation.tailrec
@@ -28,17 +27,21 @@ private[cats] class MonadGenerator extends Monad[Generator] {
   def flatMap[A, B](fa: Generator[A])(f: A => Generator[B]): Generator[B] =
     fa.flatMap(f)
 
-  // note: currently not stack safe
-  def tailRecM[A, B](a: A)(f: A => Generator[Either[A, B]]): Generator[B] =
-    new Generator[B] {
-      lazy val step = f(a)
-      def requirements: Set[Real] = step.requirements
-      def get(implicit r: RNG, n: Numeric[Real]): B =
-        step.get match {
-          case Right(b) => b
-          case Left(aa) => tailRecM(aa)(f).get
-        }
-    }
+  def tailRecM[A, B](a: A)(f: A => Generator[Either[A, B]]): Generator[B] = {
+    @tailrec
+    def run(g: Generator[Either[A, B]], r: RNG, n: Numeric[Real]): B =
+      g match {
+        case Generator.Const(_, Left(a))  => run(f(a), r, n)
+        case Generator.Const(_, Right(b)) => b
+        case Generator.From(_, fromFn) =>
+          fromFn(r, n) match {
+            case Left(a)  => run(f(a), r, n)
+            case Right(b) => b
+          }
+      }
+    val step = f(a)
+    Generator.require(step.requirements)(run(step, _, _))
+  }
 }
 
 private[cats] class MonadRandomVariable extends Monad[RandomVariable] {
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
@@ -129,12 +129,11 @@ object Generator {
 
   def traverse[K, V, W](m: Map[K, V])(
       implicit toGen: ToGenerator[V, W]
-  ): Generator[Map[K, W]] = {
+  ): Generator[Map[K, W]] =
     m.foldLeft(Generator.constant[Map[K, W]](Map.empty)) {
       case (g, (k, v)) =>
         g.zip(toGen(v)).map { case (m, w) => m.updated(k, w) }
     }
-  }
 }
 
 trait ToGenerator[-T, U] {
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
@@ -6,33 +6,55 @@ import com.stripe.rainier.sampler.RNG
 /**
   * Generator trait, for posterior predictive distributions to be forwards sampled during sampling
   */
-trait Generator[T] { self =>
+sealed trait Generator[T] { self =>
+  import Generator.{Const, From}
+
   def requirements: Set[Real]
 
   def get(implicit r: RNG, n: Numeric[Real]): T
 
-  def map[U](fn: T => U): Generator[U] = new Generator[U] {
-    val requirements: Set[Real] = self.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): U = fn(self.get)
+  private[core] def withRequirements(newReqs: Set[Real]): Generator[T] =
+    self match {
+      case Const(reqs, t)     => Const(reqs ++ newReqs, t)
+      case From(reqs, fromFn) => From(reqs ++ newReqs, fromFn)
+    }
+
+  def map[U](fn: T => U): Generator[U] = self match {
+    case Const(reqs, t)     => Const(reqs, fn(t))
+    case From(reqs, fromFn) => From(reqs, (r, n) => fn(fromFn(r, n)))
   }
 
-  def flatMap[U](fn: T => Generator[U]): Generator[U] = new Generator[U] {
-    val requirements: Set[Real] = self.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): U = fn(self.get).get
+  def flatMap[U](fn: T => Generator[U]): Generator[U] = self match {
+    case Const(reqs, t) => fn(t).withRequirements(reqs)
+    case From(reqs, fromFn) =>
+      From(reqs, { (r, n) =>
+        fn(fromFn(r, n)) match {
+          case Const(_, u)          => u
+          case From(_, innerFromFn) => innerFromFn(r, n)
+        }
+      })
   }
 
-  def zip[U](other: Generator[U]): Generator[(T, U)] = new Generator[(T, U)] {
-    val requirements: Set[Real] = self.requirements ++ other.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): (T, U) = (self.get, other.get)
+  def zip[U](other: Generator[U]): Generator[(T, U)] = {
+    val reqs: Set[Real] = self.requirements ++ other.requirements
+    (self, other) match {
+      case (Const(_, t), Const(_, u)) => Const(reqs, (t, u))
+      case (From(_, lf), From(_, rf)) =>
+        From(reqs, (r, n) => (lf(r, n), rf(r, n)))
+      case (Const(_, t), From(_, rf)) => From(reqs, (r, n) => (t, rf(r, n)))
+      case (From(_, lf), Const(_, u)) => From(reqs, (r, n) => (lf(r, n), u))
+    }
   }
 
-  def repeat(k: Real): Generator[Seq[T]] = new Generator[Seq[T]] {
-    val requirements: Set[Real] = self.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): Seq[T] =
-      0.until(n.toInt(k)).map { _ =>
-        self.get
-      }
-  }
+  def repeat(k: Real): Generator[Seq[T]] =
+    Generator.require(requirements)(self match {
+      case Const(_, u) =>
+        (_, n) =>
+          Seq.fill(n.toInt(k))(u)
+      case From(_, fromFn) =>
+        (r, n) =>
+          Seq.fill(n.toInt(k))(fromFn(r, n))
+    })
 
   private[core] def prepare(variables: Seq[Variable])(
       implicit r: RNG): Array[Double] => T = {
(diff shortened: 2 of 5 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -6,33 +6,53 @@ import com.stripe.rainier.sampler.RNG
...
+    case From(reqs, fromFn) => From(reqs, (r, n) => fn(fromFn(r, n)))
   }
 
-  def flatMap[U](fn: T => Generator[U]): Generator[U] = new Generator[U] {
-    val requirements: Set[Real] = self.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): U = fn(self.get).get
+  def flatMap[U](fn: T => Generator[U]): Generator[U] = self match {
+    case Const(reqsL, t) =>
+      fn(t) match {
+        case Const(reqsR, u)     => Const(reqsL ++ reqsR, u)
```
sritchie, line 24: This would be an actual behavior change, so let me comment. In the old implementation, only `self.requirements` propagated to the returned instance, NOT the internal requirements. If we're flatmapping two `Constant[T]` instances, however, we have access to both `reqL` and `reqR` and can take their union.

This isn't possible in the case where `self` is of type `From[T]`, so I left the behavior the same as old, and just passed `self.requirements` along.

Do we want to stay consistent, here? I don't know how requirements is used in the codebase, so it's up to y'all.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -6,33 +6,53 @@ import com.stripe.rainier.sampler.RNG
...
+      })
   }
 
-  def zip[U](other: Generator[U]): Generator[(T, U)] = new Generator[(T, U)] {
-    val requirements: Set[Real] = self.requirements ++ other.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): (T, U) = (self.get, other.get)
+  def zip[U](other: Generator[U]): Generator[(T, U)] = {
+    val reqs: Set[Real] = self.requirements ++ other.requirements
+    (self, other) match {
+      case (Const(_, t), Const(_, u)) => Const(reqs, (t, u))
```
sritchie, line 39: happy to take and implement suggestions on variable names.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -6,33 +6,53 @@ import com.stripe.rainier.sampler.RNG
...
 
-  def repeat(k: Real): Generator[Seq[T]] = new Generator[Seq[T]] {
-    val requirements: Set[Real] = self.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): Seq[T] =
-      0.until(n.toInt(k)).map { _ =>
-        self.get
-      }
-  }
+  def repeat(k: Real): Generator[Seq[T]] =
+    Generator.require(requirements)(self match {
```
sritchie, line 48: I can use `From[Seq[T]](requirements)(....)` directly here instead of `Generator.require` if you like, since we're still inside the trait. It's up to you.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -79,49 +99,40 @@ object Generator {
...
-      def get(implicit r: RNG, n: Numeric[Real]): T = fn(r, n)
-    }
+  case class Const[T](requirements: Set[Real], t: T) extends Generator[T] {
+    def get(implicit r: RNG, n: Numeric[Real]): T = t
+  }
 
-  def real(x: Real): Generator[Double] = new Generator[Double] {
-    val requirements: Set[Real] = Set(x)
-    def get(implicit r: RNG, n: Numeric[Real]) = n.toDouble(x)
+  case class From[T](requirements: Set[Real], fn: (RNG, Numeric[Real]) => T)
```
sritchie, line 106: So good!

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -79,49 +99,40 @@ object Generator {
...
 
   def traverse[T, U](seq: Seq[T])(
       implicit toGen: ToGenerator[T, U]
-  ): Generator[Seq[U]] = {
-    val asGen = seq.map(toGen(_))
-    new Generator[Seq[U]] {
-      val requirements: Set[Real] = asGen.flatMap(_.requirements).toSet
-      def get(implicit r: RNG, n: Numeric[Real]): Seq[U] = asGen.map(_.get)
+  ): Generator[Seq[U]] =
+    seq.foldLeft(Generator.constant[Seq[U]](Seq.empty)) { (g, t) =>
```
sritchie, line 126: I think this is more efficient, since we're not doing multiple passes.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -79,49 +99,40 @@ object Generator {
...
-  )(
+  def traverse[K, V, W](m: Map[K, V])(
       implicit toGen: ToGenerator[V, W]
   ): Generator[Map[K, W]] = {
-    val asGen = seq.mapValues(toGen(_))
-    new Generator[Map[K, W]] {
-      val requirements: Set[Real] = asGen.values.flatMap(_.requirements).toSet
-      def get(implicit r: RNG, n: Numeric[Real]): Map[K, W] =
-        asGen.mapValues(_.get)
+    m.foldLeft(Generator.constant[Map[K, W]](Map.empty)) {
```
sritchie, line 133: same here.

rainier-tests/src/test/scala/com/stripe/rainier/cats/suites.scala
```diff
@@ -17,8 +17,7 @@ class GeneratorSuite extends CatsSuite {
   implicit def eqGenerator[A](implicit EqA: Eq[A]): Eq[Generator[A]] =
     Eq.instance((x, y) => EqA.eqv(x.get, y.get))
 
-  checkAll("Generator[Int]",
-           MonadTests[Generator].stackUnsafeMonad[Int, Int, Int])
+  checkAll("Generator[Int]", MonadTests[Generator].monad[Int, Int, Int])
```
sritchie, line 20: voila!

sritchie (commented):
Here are some notes on my changes.

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -28,17 +27,21 @@ private[cats] class MonadGenerator extends Monad[Generator] {
...
-    new Generator[B] {
-      lazy val step = f(a)
-      def requirements: Set[Real] = step.requirements
-      def get(implicit r: RNG, n: Numeric[Real]): B =
-        step.get match {
-          case Right(b) => b
-          case Left(aa) => tailRecM(aa)(f).get
-        }
-    }
+  def tailRecM[A, B](a: A)(f: A => Generator[Either[A, B]]): Generator[B] = {
```
johnynek, line 30: I think this is a faithful translation of the original tailRecM, but I don't think the original tailRecM is correct.

Note, after the first application of `f` we are discarding the requirements. I think we need to be accumulating them in the loop.

Something like:
```scala
def run(a: A, reqs: Set[Real], r: RNG, n: Numeric[Real]): (Set[Real], B) =
  f(a) match {
    case Generator.Const(r1, Right(b)) => (reqs ++ r1, b)
    case Generator.Const(r1, Left(a)) => run(a, reqs ++ r1, r, n)
    case Generator.From(r1, fromFn) =>
      fromFn(r, n) match {
        case Right(b) => (reqs ++ r1, b)
        case Left(a) => run(a, reqs ++ r1, r, n)
      }
  }
```
But that leaves open what to do with all these accumulated requirements?

That said, looking at the original flatMap, it also discards the requirements for all but the first in the chain. That seems really strange to me....

sritchie: yeah, it's super strange to me too. I added a note to the `flatMap` implementation about that (you probably saw it). I didn't see a clear way to get the requirements up and out of the body of `From`'s function; that's the challenge that was maybe ignored in the first place, that pops up in `flatMap` of course. I just don't know enough about the reason for `requirements`.

sritchie: Funny, btw, I had that exact signature and implementation but cut it out with no mention...

sritchie: Wait, I can fix it. Hold tight. (Will delete this comment after.)

avi-stripe: I'll dig into this PR in more detail later, but to address the deal with `requirements`: the idea is that to the extent that we can anticipate `Real` objects that will be converted to double using the `Numeric[Real]` passed to `get`, we can pre-compute them more efficiently using the Real->JVM bytecode compiler; otherwise we have to fallback dynamically to a Real interpreter. So accumulating `requirements` is a very useful optimization, but semantically it's fine to discard in any case where we can't figure out how to do it.

avi-stripe:
@sritchie so great to see you contributing!

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
johnynek: We could keep them if we made them a return value of `def get(r: RNG, n: Numeric[Real]): (Set[Real], A)` would that defeat the purpose because you have to call get first?

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
avi-stripe: This way seems fine, thanks.

avi-stripe (approved):
Fantastic, sam. Thank you.

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
avi-stripe: It would yep.

sritchie: This will do it for `tailRecM`... looking again for a way to do it without calling `get` in the `flatMap` implementation.

```scala
  def tailRecM[A, B](a: A)(f: A => Generator[Either[A, B]]): Generator[B] = {
    @tailrec
    def run(g: Generator[Either[A, B]],
            reqs: Set[Real],
            r: RNG,
            n: Numeric[Real]): (Set[Real], B) =
      g match {
        case Generator.Const(r1, Left(a))  => run(f(a), reqs ++ r1, r, n)
        case Generator.Const(r1, Right(b)) => (reqs ++ r1, b)
        case Generator.From(r1, fromFn) =>
          fromFn(r, n) match {
            case Left(a)  => run(f(a), reqs ++ r1, r, n)
            case Right(b) => (reqs ++ r1, b)
          }
      }
    val step = f(a)
    Generator
      .require(step.requirements)(run(step, Set.empty, _, _))
      .flatMap { case (reqs, u) => Generator.Const(reqs, u) }
  }
```

avi-stripe: I don't know/grok `tailRecM` well enough to speak to it, but in general it'll be impossible to accumulate `requirements` across a `flatMap` in any way that's effective; this is the standard thing of not being able to statically analyze a monad (vs an applicative).

sritchie: actually, @avi-stripe, of course that is right - this just kicks it back to the `flatMap` implementation, which in my comment above I note doesn't work yet. (Because it can't work.) I'll forget it and leave the code as is.

rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
```diff
@@ -6,33 +6,53 @@ import com.stripe.rainier.sampler.RNG
...
+          case Const(_, u)          => u
+          case From(_, innerFromFn) => innerFromFn(r, n)
+        }
+      })
   }
 
-  def zip[U](other: Generator[U]): Generator[(T, U)] = new Generator[(T, U)] {
-    val requirements: Set[Real] = self.requirements ++ other.requirements
-    def get(implicit r: RNG, n: Numeric[Real]): (T, U) = (self.get, other.get)
+  def zip[U](other: Generator[U]): Generator[(T, U)] = {
```
sritchie, line 36: this is still not stack safe, which means `traverse` is not either.

sritchie:
@avi-stripe , here's one I don't understand. My change is causing an error in MY code at this place:

https://github.com/sritchie/rainier/blob/sritchie/stack_safe_generator/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala#L38

could this have something to do with the extra requirements we're now pulling through, at least in the `Const` case?

avi-stripe:
@sritchie are we shadowing the scala builtin `require` method?

sritchie:
@avi-stripe I dont' think so.

```
[error] (run-main-8) java.lang.IllegalArgumentException: requirement failed
[error] java.lang.IllegalArgumentException: requirement failed
[error] 	at scala.Predef$.require(Predef.scala:268)
[error] 	at com.stripe.rainier.core.Categorical.$anonfun$generator$3(Categorical.scala:38)
```

Gotta run for now... I'll take a look later when I get a chance, with the specific generator instance I was using.

avi-stripe:
@sritchie gotcha. so I'm sure you're right that it's related to extra requirements, and in particular, changing from the interpreter (which uses BigDecimal) to using compiled floating point ops. So we're losing precision somewhere and are no longer within that 1e-6 epsilon. It would be interesting to know how far off we are; can probably relax it to a larger eps.

sritchie:
Something major is wrong in this particular case:

```
// Math.abs(n.toDouble(cdf.last._2) - 1.0) == 0.9
[error] (run-main-c) java.lang.IllegalArgumentException: requirement failed
```

I'll reproduce with a smaller example, maybe we can track down some other bug.

avi-stripe:
@sritchie yikes. Looking forward to seeing the repro.

sritchie:
Okay, @avi-stripe, I've got my repro. I was allowing an empty sequence into the key of a map I passed to `Categorical.normalize`. As you can see below, the error occurred when I then flatMapped to `Categorical.list`, to generate items from whichever sequence the generator returned by `normalize` selected.

The `0.9` error returned was the weight I'd associated with the empty sequence.

So, this error unrelated to this PR or requirements, it turns out, and I can restructure my code to do a better job here.

I'll drop the test file here if you see some fix and want to take it... otherwise I'll call this PR complete.

Thanks!

```scala

import com.stripe.rainier.compute.{Evaluator, Real}
import com.stripe.rainier.sampler.{RNG, ScalaRNG}
import org.scalatest.FunSuite

class CategoricalTest extends FunSuite {
  implicit val rng: RNG = ScalaRNG(1527608515939L)
  implicit val evaluator: Numeric[Real] = new Evaluator(Map.empty)

  test("Categorical.list removes empty items.") {
    val expected = Seq(1, 2, 3)
    val gen = Categorical
      .normalize(
        Map(
          expected.toSeq -> 0.1,
          Seq.empty -> 0.9
        )
      )
      .flatMap(Categorical.list(_))
      .generator

    val choice = gen.get
    assert(
      expected.contains(choice),
      s"The generator returned an unexpected value - $choice - not contained in $expected."
    )
  }
}
```

avibryant:
@sritchie got it, yes, this seems unrelated but something we should protect better against. This PR is ready to merge IMO.

sritchie:
Thanks, @avibryant! When you get a chance would you mind publishing a snapshot build?

avi-stripe:
@sritchie I think we can do a quick 0.2.3 release later this week

sritchie:
Thanks @avi-stripe , that sounds great. I need it to get https://github.com/sritchie/scala-rl building on Travis..

PR #368, opened 2019-08-31
Implement cats Comonad instances for Generator and RandomVariable

sritchie:
This PR does what it says on the label! I've also added tests for each new instance, super easy due to Cats's wonderful shared laws.

I need this to abstract out some behavior in a side project... and aren't more instances just so nice?

### DIFF
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -4,14 +4,32 @@ import com.stripe.rainier.compute.Real
 import com.stripe.rainier.core.Generator
 import com.stripe.rainier.core.RandomVariable
 import com.stripe.rainier.sampler.RNG
-import _root_.cats.Monad
+import _root_.cats.{Comonad, Monad}
 
 import scala.annotation.tailrec
 
-object `package` {
+object `package` extends LowPriorityInstances {
   implicit val rainierMonadGenerator: Monad[Generator] = new MonadGenerator
   implicit val rainierMonadRandomVariable: Monad[RandomVariable] =
     new MonadRandomVariable
+}
+
+private[cats] sealed abstract class LowPriorityInstances {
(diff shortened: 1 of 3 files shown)


rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -4,16 +4,34 @@ import com.stripe.rainier.compute.Real
...
+object `package` extends LowPriorityInstances {
   implicit val rainierMonadGenerator: Monad[Generator] = new MonadGenerator
   implicit val rainierMonadRandomVariable: Monad[RandomVariable] =
     new MonadRandomVariable
 }
 
+private[cats] sealed abstract class LowPriorityInstances {
+  implicit def generatorComonad(implicit r: RNG,
+                                n: Numeric[Real]): Comonad[Generator] =
+    new MonadGenerator with Comonad[Generator] {
```
sritchie, line 20: I could ALSO have just changed the monad instance to extend `BiMonad` and added `coflatMap` and `extract` directly to the old class... but I didn't want to break binary compatibility.

rainier-scalacheck/src/main/scala/com/stripe/rainier/scalacheck/package.scala
```diff
@@ -50,4 +50,11 @@ object `package` {
     arbitrary[A].flatMap(a =>
       arbitrary[Real].flatMap(density => RandomVariable(a, density)))
 
+  implicit def cogenGenerator[A](implicit CA: Cogen[A],
```
sritchie, line 53: The Comonad tests needed these. I'm not familiar enough with the stuff above to see if this new code is helpful with existing stuff.

sritchie (commented):
Just some notes on my implementation.

avibryant:
This LGTM, though it wouldn't hurt for @erik-stripe or @oscar-stripe to take a look also.

PR #369, opened 2019-09-03
Bump version to rc3 and fix some types

metcalf:
In the process of pulling in the latest Rainier I noticed some small issues with our latest changes that are reflected here. Upon landing this I'll publish `0.2.3-rc3` internally. Note that I already published `0.2.3-rc2` without the fixes in this PR so I think we have to skip an RC number.


avibryant:
LGTM

PR #370, opened 2019-09-04
add fromSet to Categorical.scala

sritchie:
This PR adds a function to generate a `Categorical` distribution from a set of items.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
@@ -90,6 +90,11 @@ object Categorical {
       Real(l.size)
     })
 
+  def fromSet[T](ts: Set[T]): Categorical[T] =
+    normalize(
+      ts.foldLeft(Map.empty[T, Real])((m, t) => m.updated(t, Real.one))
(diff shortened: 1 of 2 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -90,6 +90,11 @@ object Categorical {
       Real(l.size)
     })
 
+  def fromSet[T](ts: Set[T]): Categorical[T] =
```
avi-stripe: Any reason to prefer this over something like 
```
val p = Real.one / ts.size
ts.map{t => t -> p}.toMap
```
?

sritchie: nope, just wanted to reuse normalize vs "re-implementing", even though it's easy. This is good, I'll fix!

sritchie: Ah, I did use the `foldLeft` to avoid two passes over the set, as `toMap` would have done (I think). I'm keeping that, but I took your `val p` suggestion.

PR #371, opened 2019-09-04
Categorical Monad implementation

sritchie:
This PR adds a `Monad[Categorical]` instance. I managed to get a stack safe implementation!

I also moved the Eq instances into the cats package, so I could get access to them outside Rainier.

It might make sense to add an implicit value for the BigDecimal epsilon with a default, so the user can override.

I also made `Generator` covariant in `T`, which required a few tweaks to how the scalacheck `Gen` and `Arbitrary` instances are defined. Their behavior is the same as before, of course.

I had to write some new testing components, and would love comments there... but this should be solid, thanks to the cats testing functions.

### DIFF
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -17,7 +17,7 @@ import scala.collection.immutable.Queue
 
 object `package` extends LowPriorityInstances with EqInstances {
   implicit val rainierMonadCategorical: Monad[Categorical] =
-    new MonadCategorical
+    MonadCategorical
   implicit val rainierMonadGenerator: Monad[Generator] = new MonadGenerator
   implicit val rainierMonadRandomVariable: Monad[RandomVariable] =
     new MonadRandomVariable
@@ -41,7 +41,7 @@ private[cats] sealed abstract class LowPriorityInstances {
     }
 }
 
-private[cats] class MonadCategorical extends Monad[Categorical] {
+private[cats] object MonadCategorical extends Monad[Categorical] {
   def pure[A](x: A): Categorical[A] = Categorical(Map(x -> Real.one))
 
   override def map[A, B](fa: Categorical[A])(f: A => B): Categorical[B] =
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -131,11 +131,13 @@ private[cats] trait EqInstances {
 
   def eqBigDecimal(epsilon: Double): Eq[BigDecimal] =
     Eq.instance { (left, right) =>
-      ((left - right) / left) < epsilon && ((left - right) / right) < epsilon
+      if (right == BigDecimal(0.0))
+        left.abs < epsilon
+      else ((left - right).abs / right.abs) < epsilon
     }
 
-  implicit val eqReal: Eq[Real] = {
-    val bde = eqBigDecimal(1e-6)
+  def eqReal(epsilon: Double): Eq[Real] = {
+    val bde = eqBigDecimal(epsilon)
     Eq.instance { (left, right) =>
       (left, right) match {
         case (Infinity, Infinity) | (NegInfinity, NegInfinity) => true
@@ -144,6 +146,8 @@ private[cats] trait EqInstances {
       }
     }
   }
+
+  implicit val defaultEqReal: Eq[Real] = eqReal(1e-6)
 
   implicit def eqCategorical[A: Eq]: Eq[Categorical[A]] = {
     implicit val mapEq: Eq[Map[A, Real]] = catsKernelStdEqForMap[A, Real]
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -2,7 +2,12 @@ package com.stripe.rainier
 package cats
 
 import com.stripe.rainier.compute.{Constant, Infinity, NegInfinity, Real}
-import com.stripe.rainier.core.{Categorical, Generator, RandomVariable}
+import com.stripe.rainier.core.{
+  updateMap,
+  Categorical,
+  Generator,
+  RandomVariable
+}
 import com.stripe.rainier.sampler.RNG
 import _root_.cats.{Eq, Monad}
 import _root_.cats.kernel.instances.map._
@@ -44,7 +49,7 @@ private[cats] class MonadCategorical extends Monad[Categorical] {
           case (Left(a), r) =>
             run(acc, queue.tail ++ f(a).pmf.mapValues(_ * r))
           case (Right(b), r) =>
-            run(acc.updated(b, acc.getOrElse(b, Real.zero) + r), queue.tail)
+            run(updateMap(acc, b, r)(Real.zero)(_ + _), queue.tail)
         }
       }
     val pmf = run(Map.empty, f(a).pmf.to[Queue])
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -142,7 +142,10 @@ private[cats] trait EqInstances {
       (left, right) match {
         case (Infinity, Infinity) | (NegInfinity, NegInfinity) => true
         case (Constant(a), Constant(b))                        => bde.eqv(a, b)
-        case _                                                 => false
+        // TODO - the Eq instance returned here doesn't test for full
+        // DAG equality; this final case needs to be expanded to
+        // compare the various NonConstant options.
+        case (a, b) => a == b
       }
     }
   }
@@ -160,5 +163,8 @@ private[cats] trait EqInstances {
       n: Numeric[Real]
   ): Eq[Generator[A]] = Eq.by(_.get)
 
+  // TODO - Comparing RandomVariable instances by value alone isn't
+  // sound; a proper Eq instance needs to take density into account as
+  // well.
   implicit def eqRandomVariable[A: Eq]: Eq[RandomVariable[A]] = Eq.by(_.value)
 }
--- a/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
+++ b/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
@@ -1,14 +1,16 @@
-package com.stripe.rainier.cats
+package com.stripe.rainier
+package cats
 
-import com.stripe.rainier.compute.Real
+import com.stripe.rainier.compute.{Constant, Infinity, NegInfinity, Real}
 import com.stripe.rainier.core.{Categorical, Generator, RandomVariable}
 import com.stripe.rainier.sampler.RNG
-import _root_.cats.Monad
+import _root_.cats.{Eq, Monad}
+import _root_.cats.kernel.instances.map._
 
 import scala.annotation.tailrec
 import scala.collection.immutable.Queue
 
-object `package` {
+object `package` extends EqInstances {
   implicit val rainierMonadCategorical: Monad[Categorical] =
     new MonadCategorical
   implicit val rainierMonadGenerator: Monad[Generator] = new MonadGenerator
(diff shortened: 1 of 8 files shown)


avi-stripe:
There's no full DAG-equality implementation for `Real` yet, but that would be a welcome contribution.

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -1,19 +1,54 @@
...
+
+  override def map[A, B](fa: Categorical[A])(f: A => B): Categorical[B] =
+    fa.map(f)
+
+  override def product[A, B](
+      fa: Categorical[A],
+      fb: Categorical[B]
+  ): Categorical[(A, B)] = fa.zip(fb)
+
+  override def flatMap[A, B](fa: Categorical[A])(
```
sritchie, line 55: Gotta love that stack safety!!

rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
```diff
@@ -114,9 +115,9 @@ object Real {
   val negInfinity: Real = NegInfinity
 }
 
-final private case class Constant(value: BigDecimal) extends Real
-final private object Infinity extends Real
-final private object NegInfinity extends Real
+final private[rainier] case class Constant(value: BigDecimal) extends Real
```
sritchie, line 118: I needed to access these in the tests to write the `Eq[Real]`  instance.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -25,6 +25,12 @@ final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T] {
         .map { case (u, ups) => u -> Real.sum(ups.map(_._2)) }
         .toMap)
 
+  def zip[U](other: Categorical[U]): Categorical[(T, U)] =
```
sritchie, line 33: We can actually implement this more efficiently in the monad than the default double-flatmap call, since we know each pair can only show up once in the final map.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -126,9 +132,9 @@ final case class Multinomial[T](pmf: Map[T, Real], k: Real)
 }
 
 object Multinomial {
-  def optional[T](pmf: Map[T,Real], k: Real): Multinomial[Option[T]] = {
+  def optional[T](pmf: Map[T, Real], k: Real): Multinomial[Option[T]] = {
```
sritchie: scalafmt butting its head in.

rainier-scalacheck/src/main/scala/com/stripe/rainier/scalacheck/package.scala
```diff
@@ -1,37 +1,27 @@
...
-import org.scalacheck.Arbitrary.arbitrary
-import org.scalacheck.Cogen
-import org.scalacheck.Gen
+import com.stripe.rainier.sampler.{RNG, ScalaRNG}
+import org.scalacheck.{Arbitrary, Cogen, Gen}
 
 object `package` {
+  import Arbitrary.arbitrary
 
-  implicit val arbitraryReal: Arbitrary[Real] =
```
sritchie, line 16: I just grouped these with the other arbitrary instances.

rainier-scalacheck/src/main/scala/com/stripe/rainier/scalacheck/package.scala
```diff
@@ -1,37 +1,27 @@
...
-    Arbitrary(arbitrary[Double].map(Real(_)))
-
-  implicit def arbitraryGenerator[A: Arbitrary]: Arbitrary[Generator[A]] =
-    Arbitrary(genGenerator)
+  implicit val cogenRNG: Cogen[RNG] =
```
sritchie: I moved this out of `genGenerator`, thinking I would need it... I don't, but it's useful outside, yeah?

rainier-tests/src/test/scala/com/stripe/rainier/cats/suites.scala
```diff
@@ -1,13 +1,45 @@
-package com.stripe.rainier.cats
+package com.stripe.rainier
```
sritchie, line 1: I did this so that I could get access to the `rainier` scope that the various `Real` instances were defined in.

rainier-tests/src/test/scala/com/stripe/rainier/cats/suites.scala
```diff
@@ -1,13 +1,45 @@
...
+import _root_.cats.kernel.instances.MapInstances
+import _root_.cats.tests.CatsSuite
+import _root_.cats.laws.discipline._
+
+class CategoricalSuite extends CatsSuite with MapInstances {
+
+  implicit val rng: RNG = RNG.default
+  implicit val evaluator: Evaluator = new Evaluator(Map.empty)
+
+  def eqBigDecimal(epsilon: Double): Eq[BigDecimal] =
```
sritchie: HUGE numbers that were percentage wise very close were screwing up my tests.

rainier-tests/src/test/scala/com/stripe/rainier/cats/suites.scala
```diff
@@ -1,13 +1,45 @@
...
+      }
+    }
+  }
+
+  implicit def eqCategorical[A: Eq]: Eq[Categorical[A]] = {
+    implicit val mapEq: Eq[Map[A, Real]] = catsKernelStdEqForMap[A, Real]
+    Eq.instance((x, y) => Eq.eqv(x.pmf, y.pmf)(mapEq))
+  }
+
+  checkAll("Categorical[Int]", MonadTests[Categorical].monad[Int, Int, Int])
```
sritchie, line 13: And we have a monad!

sritchie (commented):
Some comments on my implementation.

sritchie:
@avi-stripe @avibryant, this is now ready to go!

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -1,19 +1,54 @@
...
+  def tailRecM[A, B](a: A)(
+      f: A => Categorical[Either[A, B]]): Categorical[B] = {
+
+    @tailrec
+    def run(acc: Map[B, Real],
+            pairs: List[(Either[A, B], Real)]): Map[B, Real] = {
+      pairs match {
+        case Nil => acc
+        case (Left(a), r) :: xs =>
+          run(acc, xs ++ f(a).pmf.mapValues(_ * r).toList)
```
johnynek: 1. I'd use `xs ::: f(a).pmf.mapValues(_ * r).toList` so we don't use the `Iterable` method and we use the List specific version, may not matter.
2. more concerning, this can be `O(N^2)` since it is appending on the right. One way to solve this would be to use a `Queue[(Either[A, B], Real)]`. instead of List.

sritchie: I like the suggestion - I've switched the implementation to use a `Queue`. Thanks!

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -65,3 +103,35 @@ private[cats] class MonadRandomVariable extends Monad[RandomVariable] {
       case Right(b) => RandomVariable(b)
     }
 }
+
+private[cats] trait EqInstances {
```
sritchie, line 130: I moved these into `rainier-cats` so that I could get access to them when writing my own tests outside the repo.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -9,21 +10,33 @@ import com.stripe.rainier.compute._
   */
 final case class Categorical[T](pmf: Map[T, Real]) extends Distribution[T] {
   self =>
+
   def map[U](fn: T => U): Categorical[U] =
-    flatMap { t =>
-      Categorical(Map(fn(t) -> Real.one))
-    }
+    Categorical(
```
sritchie, line 15: this is a bit more efficient.

rainier-core/src/main/scala/com/stripe/rainier/core/Categorical.scala
```diff
@@ -9,21 +10,33 @@ import com.stripe.rainier.compute._
...
     Categorical(
-      pmf.toList
-        .flatMap {
-          case (t, p) =>
-            fn(t).pmf.toList.map { case (u, p2) => (u, p * p2) }
-        }
-        .groupBy(_._1)
-        .map { case (u, ups) => u -> Real.sum(ups.map(_._2)) }
-        .toMap)
+      (for {
```
sritchie, line 24: This implementation is more efficient, as we take fewer passes over each of the `pmf` instances.

rainier-core/src/main/scala/com/stripe/rainier/core/package.scala
```diff
@@ -0,0 +1,8 @@
+package com.stripe.rainier
+
+package object core {
```
sritchie, line 3: shared logic from `map`, `flatMap` and `tailRecM`.

sritchie (commented):
Okay, @avi-stripe, done for good with this PR.

avibryant:
Thanks @sritchie. I'm going to need a bit more time to review this, sorry (super busy until Monday). Looks like there are some conflicts to resolve now that we've merged some of your other PRs in the meantime.

sritchie:
Not a problem at all, @avibryant - thanks for the other merges! I've fixed the merge conflicts here, so this PR is ready to go when you get some time.

sritchie:
I left this comment on #378 about why I've added the final commit here, which makes `Generator` covariant in `T`. https://github.com/stripe/rainier/pull/378#issuecomment-529430543

sritchie:
Hey @avibryant, I pushed a change to the `Eq` implementation for `BigDecimal` after realizing that it would explode if either side was completely equal to zero. No other changes, here.

The new one is better:

```scala
  def eqBigDecimal(epsilon: Double): Eq[BigDecimal] =
    Eq.instance { (left, right) =>
      if (right == BigDecimal(0.0))
        left.abs < epsilon
      else ((left - right).abs / right.abs) < epsilon
    }
```

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -83,3 +126,39 @@ private[cats] class MonadRandomVariable extends Monad[RandomVariable] {
...
+      else ((left - right).abs / right.abs) < epsilon
+    }
+
+  def eqReal(epsilon: Double): Eq[Real] = {
+    val bde = eqBigDecimal(epsilon)
+    Eq.instance { (left, right) =>
+      (left, right) match {
+        case (Infinity, Infinity) | (NegInfinity, NegInfinity) => true
+        case (Constant(a), Constant(b))                        => bde.eqv(a, b)
+        case _                                                 => false
```
avi-stripe: Even without going to a full DAG-equality, what about adding another case for when they are identically the same object?

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
```diff
@@ -83,3 +126,39 @@ private[cats] class MonadRandomVariable extends Monad[RandomVariable] {
...
+    Eq.by(_.pmf)
+  }
+
+  implicit def eqGenerator[A](
+      implicit eqA: Eq[A],
+      r: RNG,
+      n: Numeric[Real]
+  ): Eq[Generator[A]] = Eq.by(_.get)
+
+  implicit def eqRandomVariable[A: Eq]: Eq[RandomVariable[A]] = Eq.by(_.value)
```
avi-stripe, line 169: This makes me cringe a bit - it's fine if it narrowly serves your needs but two RVs with equal values and different densities are definitely not equal. I'd suggest at least warning about that in comments.

avi-stripe (commented):
Sorry for the long delay, @sritchie. This generally looks good, but I'm uneasy about the limited Eq instances, especially for RV. I think it's ok to just comment them with TODO/WARNING right now but let's be mindful of the need for something better.

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
sritchie: Warning added - my mistake, I don't get how to write a proper `Eq` method that compares densities yet. Happy to do so with some guidance, perhaps in the next PR?

rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala
sritchie: Boom, done. Let me add a TODO as well, as you've suggested.

sritchie:
Okay @avi-stripe , this should now be good to go - I've added the comments you requested as well as that case testing object equality.

avi-stripe:
Thanks @sritchie!

PR #376, opened 2019-09-06
More Cats instances

sritchie:
This is a draft PR that adds instances for:

- `Group[Real]`
- `Monoid[RandomVariable[A]]`
- `Monoid[Generator[A]]`

Just experimenting with how much we can do here.

I'm running into this on `RealGroup`'s tests:

```
[info] - Real.group.collect0 *** FAILED ***
[info]   ArithmeticException was thrown during property evaluation.
[info]     Message: Division undefined
[info]     Occurred when passed generated values (
[info]       arg0 = Constant(-1.1327590390904958E-124)
[info]     )
[info] - Real.group.combine all *** FAILED ***
[info]   ArithmeticException was thrown during property evaluation.
[info]     Message: Division undefined
[info]     Occurred when passed generated values (
[info]       arg0 = Vector()
[info]     )
```

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
@@ -42,7 +42,7 @@ object Events {
       }
     })
 
-  def simulate[X, Y, D](seq: Seq[X], fn: Fn[X,D])(
+  def simulate[X, Y, D](seq: Seq[X], fn: Fn[X, D])(
       implicit tg: ToGenerator[D, Y]): Generator[Seq[(X, Y)]] =
     simulate(seq)(fn(_))
 }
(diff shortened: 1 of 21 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
```diff
@@ -1,16 +1,26 @@
 package com.stripe.rainier.core
```
sritchie: This is a scalafmt change that keeps sneaking in on my branch.

sritchie:
Okay @avi-stripe, this should be my last one for a bit! I played around with adding a monad transformer for Generator, but I don't think it helps anything I'm doing. It's too fun to play around with this stuff.

avi-stripe:
Awesome. Thank you!

PR #377, opened 2019-09-06
add contramap implementation to ToReal

sritchie:
This PR adds a `contramap` implementation; this lets me get instances of `ToReal[B]` for types where I can extract something numeric, say, but for which I can't actually define a `Numeric[B]` instance.

I'm using this with algebird's `AveragedValue`. I'd like to be able to have the `ToReal` implicit handle the conversion rather than explicitly taking a function from `T => Double`, in this case. With this PR I can write:

```scala
  implicit val avToReal: ToReal[AveragedValue] =
    implicitly[ToReal[Double]].contramap(_.value)
```


avibryant:
LGTM

PR #378, opened 2019-09-08
Make Generator covariant in T

sritchie:
This PR adds a `+` to the type parameter on Generator... I've got a few cases where I'd like to use items of type `Generator[MapState[A, R]]` as input to functions that call for `Generator[State[A, R]]`, and this is the solution.

I don't THINK anything else needs to change, here... let's see how the tests go.

(I goofed my branch name, as you can see. I can't keep these two concept (name)s straight...)

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
@@ -6,7 +6,7 @@ import com.stripe.rainier.sampler.RNG
 /**
   * Generator trait, for posterior predictive distributions to be forwards sampled during sampling
   */
-sealed trait Generator[T] { self =>
+sealed trait Generator[+T] { self =>
   import Generator.{Const, From}
 
   def requirements: Set[Real]


avibryant:
Fine once the tests pass - looks like there's a compilation error on 2.11.

sritchie:
I don't totally understand WHY this was the fix, but here's the fix:

https://github.com/stripe/rainier/pull/371/commits/a57731e726622fa94a99884685e0d7e87a753df8

I've added this to the #371 since I was already changing `scalacheck/package.scala` there. The fix was to modify the scalacheck `Gen` instances so that they didn't try to resolve their components as implicit `Arbitrary` instances; now, all `Gen` instances take other `Gen` instances directly, and take no implicits at all.

The implicit generators are supplied when we define the `Arbitrary` instances at the bottom. This resolves the diverging implicit expansion.

I'm happy to back those changes off and move them back to this PR once (and if!) you like the categorical monad and decide to merge it. Just let me know what works best.

sritchie:
Closing since this is absorbed into the categorical monad P.R.

PR #379, opened 2019-09-12
tweak the Events API

avi-stripe:
These are changes that were suggested by working on the Ch 5.1 notebook. Specifically:

* I've brought back the convenience methods that separate out a Seq[X] and Seq[Y]; these turn out to be a lot more readable IMO.
* `Events.simulate` has been moved to an object method that works with any X=>Generator[T], not just X=>Distribution[Y]. This is useful when you want to visualize counterfactual or predictive posteriors on functions that don't match the data's (X,Y) format, ie when you want (X,mu) in a regression.
* `observe` now just always returns the bare `X => Distribution[Y]` function rather than a case class wrapper


andrew-stripe (approved):
Looks good modulo Fn support (or at least a TODO for it)

PR #380, opened 2019-09-17
sort whisker plots by value

avi-stripe:
Sorting by key is much less useful than sorting by value.


avi-stripe:
going to self-merge in order to get an internal RC out cc @andrew-stripe

andrew-stripe:
👍

PR #383, opened 2019-09-24
WAIC

avi-stripe:
An apparently-working implementation of WAIC.


rainier-core/src/main/scala/com/stripe/rainier/compute/Coefficients.scala
```diff
@@ -16,8 +16,13 @@ sealed trait Coefficients {
 }
 
 object Coefficients {
-  def apply(pair: (NonConstant, BigDecimal)): Coefficients =
-    One(pair._1, pair._2)
+  def apply(pair: (NonConstant, BigDecimal)): Coefficients = {
```
avi-stripe: This is an obscure bug that this PR turned up.

rainier-core/src/main/scala/com/stripe/rainier/core/Events.scala
```diff
@@ -42,7 +42,7 @@ object Events {
       }
     })
 
-  def simulate[X, Y, D](seq: Seq[X], fn: Fn[X,D])(
+  def simulate[X, Y, D](seq: Seq[X], fn: Fn[X, D])(
```
avi-stripe: unrelated formatting

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -80,6 +81,17 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       }
   }
 
+  def waic(sampler: Sampler,
```
avi-stripe, line 83: we have too many variations of sampling methods that have these same basic parameters, this is crying out for a refactoring

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
   def fit[L, T](l: L, seq: Seq[T])(
       implicit toLH: ToLikelihood[L, T]): RandomVariable[Unit] =
     toLH(l).fit(seq)
+
+  private[core] def waic(samples: List[Array[Double]],
```
avi-stripe: potentially this should be broken out into a WAIC object?

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+      val cf = Compiler.default.compile(
+        target.placeholderVariables ++ variables,
+        List(("logLikelihood", target.real)))
+      val globalBuf = new Array[Double](cf.numGlobals)
+      val inputBuf = new Array[Double](cf.numInputs)
+      val data = target.placeholderVariables.map { v =>
+        target.placeholders(v)
+      }.toArray
+      val sampleArray = samples.toArray
+      var i = 0
```
avi-stripe: trying to be allocation-free from this point on...

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+                   cf: CompiledFunction,
+                   inputBuf: Array[Double],
+                   globalsBuf: Array[Double],
+                   data: Array[Array[Double]],
+                   i: Int): Double = {
+    val nInputs = inputBuf.size
+    val nDataInputs = data.size
+    val nSampleInputs = nInputs - nDataInputs
+    val nSamples = samples.size
+    val outputs = new Array[Double](nSamples)
```
avi-stripe: ... except oops, this is an allocation. will fix later.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+      j = 0
+      while (j < nSampleInputs) {
+        inputBuf(nDataInputs + j) = sample(j)
+        j += 1
+      }
+      outputs(i) = cf.output(inputBuf, globalsBuf, 0)
+      k += 1
+    }
+
+    val max = outputs.max
```
avi-stripe: the "log-sum-exp trick"

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+    while (k < nSamples) {
+      expSum += Math.exp(outputs(i) - max)
+      k += 1
+    }
+    val lppd = Math.log(expSum) + max - Math.log(nSamples.toDouble)
+    val pWAIC = variance(outputs)
+    lppd - pWAIC
+  }
+
+  private def variance(values: Array[Double]): Double = {
```
avi-stripe: welford's online variance algorithm, from wikipedia

rainier-tests/src/test/scala/com/stripe/rainier/core/WAICTest.scala
```diff
@@ -0,0 +1,75 @@
+package com.stripe.rainier.core
+
+import com.stripe.rainier.sampler._
+import org.scalatest.FunSuite
+
+class WAICTest extends FunSuite {
+  implicit val rng = RNG.default
+  val cars = List(
```
avi-stripe, line 8: the cars dataset from R

rainier-tests/src/test/scala/com/stripe/rainier/core/WAICTest.scala
```diff
@@ -0,0 +1,75 @@
...
+      b <- Normal(0, 10).param
+      sigma <- Uniform(0, 30).param
+      _ <- Events.observe(cars) { s =>
+        val mu = a + b * s
+        Normal(mu, sigma)
+      }
+    } yield (a, b)
+
+    val w = m.waic(HMC(5), 100000, 10000, 1)
+    assert(w >= 400.0 && w <= 450.0)
```
avi-stripe: per Statistical Rethinking ch 6

avi-stripe (commented):
Left a lot of review comments that should probably be comment comments :)

.vscode/settings.json
```diff
@@ -1,3 +1,2 @@
 {
-    "metals.serverVersion": "0.7.0"
```
andrew-stripe, line 2: Intentional?

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
andrew-stripe: That seems reasonable, after sampling this seems separable from the RandomVariable.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
andrew-stripe: Agreed. I think having `waic` live on some `Sample` type would be much better. 

Meanwhile, should there be a version of the `waic` method that takes a list of samples? I imagine it'll be pretty common to both want the sample from the posterior and the WAIC.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
andrew-stripe: This should probably be a comment and the exception a TODO comment

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
andrew-stripe: Needs to be an `outputBuf` preallocated in the caller I guess?

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+    }
+    var k = 0
+    while (k < nSamples) {
+      val sample = samples(k)
+      j = 0
+      while (j < nSampleInputs) {
+        inputBuf(nDataInputs + j) = sample(j)
+        j += 1
+      }
+      outputs(i) = cf.output(inputBuf, globalsBuf, 0)
```
andrew-stripe: I assume the `0` here means the CF only has a single output we care about?

andrew-stripe: Is this supposed to be `outputs(k)`? Otherwise it's overwriting the same value on every turn of the loop.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+    }
+
+    val max = outputs.max
+    var expSum = 0.0
+    k = 0
+    while (k < nSamples) {
+      expSum += Math.exp(outputs(i) - max)
+      k += 1
+    }
+    val lppd = Math.log(expSum) + max - Math.log(nSamples.toDouble)
```
andrew-stripe: Ahh, took me a few minutes to realize that you moved the division for the average into a subtraction in log space. Maybe worth a comment.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
andrew-stripe: I wasn't quite sure why this was necessary till I remembered that outputs contains the log likelihood so we need to exp it before summing. Maybe worth a comment though may be obvious to someone more familiar with this code.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -226,4 +238,85 @@ object RandomVariable {
...
+                   data: Array[Array[Double]],
+                   i: Int): Double = {
+    val nInputs = inputBuf.size
+    val nDataInputs = data.size
+    val nSampleInputs = nInputs - nDataInputs
+    val nSamples = samples.size
+    val outputs = new Array[Double](nSamples)
+    var j = 0
+    while (j < nDataInputs) {
+      inputBuf(j) = data(j)(i)
```
andrew-stripe: I don't know the details of inputs/data/etc... so I can't review this in as much detail for correctness.

rainier-tests/src/test/scala/com/stripe/rainier/core/WAICTest.scala
andrew-stripe: If you split WAIC into a separate object I think you'll be able to pass an array of samples that let you test an exact value.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
avi-stripe: > a version of the waic method that takes a list of samples

The problem is that what the waic method needs is the samples of the raw parameter valus, whereas when you `sample()` you're getting some transformed output/prediction. This is why I think we need some kind of `Sample` or `Posterior` object that encapuslates both...

.vscode/settings.json
avi-stripe: yes, though unrelated

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
avi-stripe: Nice catch!

rainier-core/src/main/scala/com/stripe/rainier/core/WAIC.scala
```diff
@@ -0,0 +1,112 @@
+package com.stripe.rainier.core
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.ir._
+
+case class WAIC(lppd: Double, pWAIC: Double) {
```
oscar-stripe, line 7: want to add a comment to explain this for the uninitiated what `WAIC` stands for maybe with a link to read more?

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -80,6 +80,19 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
       }
   }
 
+  def waic(sampler: Sampler,
```
andrew-stripe, line 83: Maybe worth adding a no-args convenience version of this to mirror `sample`.

rainier-core/src/main/scala/com/stripe/rainier/repl/package.scala
```diff
@@ -177,5 +178,36 @@ package object repl {
   private def leftPad(s: String, len: Int, elem: Char): String =
     s.reverse.padTo(len, elem).reverse
 
+  def compare(models: (String, RandomVariable[_])*): Unit = {
+    //hardcoding the sampling parameters for now; we really need to refactor the sampling API to improve this
```
andrew-stripe: It's pretty easy to add the version that takes sampler parameters to mirror `sample`. Maybe worth it given how often we've had to tune the sampler in the book.

andrew-stripe (approved):
Two minor comments, looks good when tests are passing.

avi-stripe:
@andrew-stripe thanks, I've added the variants to `waic` and `compare`.

PR #384, opened 2019-09-25
Add theme to the shade interface to allow customizing fill

andrew-stripe:
Using the plot theme for the color means we can't have multiple shaded regions with different colors. With this PR I was able to generate:
image

### DIFF
--- a/rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
+++ b/rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
@@ -188,7 +188,8 @@ object Jupyter {
   }
 
   def shade[M, N](intervals: Seq[(M, (N, N))])(implicit mNum: Numeric[M],
-                                               nNum: Numeric[N]): Plot = {
+                                               nNum: Numeric[N],
+                                               theme: Theme): Plot = {
     val doubleTriples = intervals
       .map {
         case (m, (n1, n2)) =>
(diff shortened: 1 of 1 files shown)


rainier-plot/src/main/scala/com/stripe/rainier/plot/Jupyter.scala
```diff
@@ -205,7 +206,7 @@ object Jupyter {
       Bounds(minX, maxX),
       Bounds(minY, maxY),
       new PlotRenderer {
-        def render(plot: Plot, plotExtent: Extent)(implicit theme: Theme) = {
+        def render(plot: Plot, plotExtent: Extent)(implicit _theme: Theme) = {
```
avi-stripe, line 209: This seems a little bit hacky to me - I feel like the intent for evilplot is to have a consistent theme throughout that is deferred until render time, and here you're ignoring the provided theme in favor of a "hardcoded" one. Seems better to me to have an explicit color parameter (that is optional and defaults to the theme's).

andrew-stripe: Are you sure that's the intent of the EvilPlot API? Our `shade` feels analogous to something like `ScatterPlot.apply` or `BarChart.apply` both of which take an implicit theme. I believe the intent is that the theme passed to each plot configures that plot and things like the `pointRenderer` optional argument are only needed if you want to customize per-point.

avi-stripe: Ok, I buy that.

avi-stripe (approved):

PR #385, opened 2019-09-25
Provide multiple themes for Jupyter

andrew-stripe:
This introduces a PlotThemes object inside Jupyter for us to create different themes. Note that this also updates the themes to customize the color on bar charts. That'll change the blue color a bit on existing charts to match scatters.

I needed the separate blue and gray themes to create plots like these:
image
image


avi-stripe:
LGTM

PR #386, opened 2019-09-25
coef and coeftab from Statistical Rethinking

andrew-stripe:
Note that we don't have access to the number of observations but I think that's reasonable to omit.

image


avi-stripe:
LGTM

PR #388, opened 2019-09-27
Coefficient(x, 0.0) should be Empty

avi-stripe:
This is a minor fix to `Coefficient` but it exposes a problem in `distribute` when all the terms cancel and we're left with a `Constant` not a `Line`.


avi-stripe:
Most of what this does now is avoids creating intermediate `Line` objects during the eager optimization phase, which (having fixed this bug) might now have illegally-empty coefficients objects, and instead use simple tuples of (ax,b) for intermediate state.

andrew-stripe (approved):

PR #392, opened 2019-10-07
Add a "vector" and "fromSet" method to Generator

sritchie:
These were two convenience methods I found useful in ScalaRL... not sure if they're too simple to include. I'm submitting them just in case.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Generator.scala
@@ -120,6 +120,10 @@ object Generator {
   def require[T](reqs: Set[Real])(fn: (RNG, Numeric[Real]) => T): Generator[T] =
     From(reqs, fn)
 
+  def vector[T](v: Vector[T]) = from((r, _) => v(r.int(v.size)))
+
+  def fromSet[T](items: Set[T]) = vector(items.to[Vector])
+
   def traverse[T, U](seq: Seq[T])(
       implicit toGen: ToGenerator[T, U]
   ): Generator[Seq[U]] =


avi-stripe:
Thanks!

avi-stripe:
@sritchie actually thinking about this more: normally I would do `Categorical.fromSet(set).generator` if I wanted this, which feels more in keeping with the rest of the API. Is there a reason you didn't want to do it that way?

sritchie:
Hey @avi-stripe, I'll have to put together some clearer benchmarks, but I found in ScalaRL that, for my basic use case of all-constant `Real`s, the `Real` / `BigDecimal` implementation of `Categorical` was very slow; my goal here was to keep sampling fast and simple in this case, where all we really need is a draw from a standard uniform distribution.

I'm probably missing some obvious way to use Rainier more effectively. I've been focusing on the reinforcement learning chapters that deal with states and policies where you're dealing with a discrete set of actions, and you have access to:

1. the weight, or probability, that the current policy assigns to each action, and
2. the probability that the state implementation will return some `(Reward, NewState)` pair given an action.

The core thing that needed to be fast was this expected value calculation:

https://github.com/sritchie/scala-rl/blob/develop/scala-rl-core/src/main/scala/io/samritchie/rl/util/ExpectedValue.scala#L22

I was originally staying in `Real`, and returning a `Value[Real]` here, then using an `Ordering[Real]` instance I built to figure out the set of actions associated with the maximum expected value. Staying completely in `Real` was very slow, so for now I'm using a simpler `Categorical` implementation that uses `Double` instead of `Real`:

https://github.com/sritchie/scala-rl/blob/develop/scala-rl-core/src/main/scala/io/samritchie/rl/Cat.scala#L19

which lets me beat the Python benchmarks, thank god. The slight tweak I made to that version is that I store `List[(T, Double)]` instead of `Map[T, Double]`, which lets me make the class covariant in `T`.

It's very possible that by using the `ExpectedValue[Categorical]` instance I linked to above:

https://github.com/sritchie/scala-rl/blob/develop/scala-rl-core/src/main/scala/io/samritchie/rl/util/ExpectedValue.scala#L20

which takes a `Numeric[Real]`, that the caching in `Evaluator` would make things fast again.

In the slowest case, I wasn't using an `Evaluator`, but just using the `Real` values of `Categorical` directly to calculate expected values (using some function of `A => Real`, and then comparing the `Real` instances)

Hope that explains the use case - you're right that conceptually calling `Categorical.fromSet(set).generator` is identical, and I should probably write a benchmark first.

avi-stripe:
@sritchie that's helpful. As a quick practical suggestion: how about changing `Categorical` to be a sealed trait and include an optimized implementation for this uniform case?

As a higher level observation, and forgive me if most/all of this is context you already have: the real investment in performance for Rainier has been for MCMC sampling approaches, where you have a `Real` that represents the log of the joint density p(θ)p(x|θ) - which per bayes, is proportional to p(θ|x) - and you have a tight loop that is just evaluating that (and maybe its gradient) for many different choices of θ. The `RandomVariable` monad is all about constructing those density functions, once, and then compiling them to be efficiently evaluated 100s of thousands of times.

The intended use of `Generator` was then to do any stochastic prediction of new observations given the sampled θ; but this generally only needs to happen a few thousand times in that context, so is less performance sensitive.

From what I understand, you're not really using RandomVariable, or sampling from the density functions it produces at all, but are just using Generator as an unconditioned probability monad. We definitely have some work in front of us if we want to optimize for that use case.

sritchie:
I do expect to use Rainier as it's intended once I get to the end of the project, where I'll build in the ability for a reinforcement learning agent to keep an internal model of the environment it's interacting with; the examples in the textbook are all "assume the state will return exactly what it did last time", of course because it's too complicated to present a full model-building exercise on top of everything else they're going through.

My hope is that with Rainier, I can build much more rich models, and take advantage of the performance boosts you've worked hard on.

I like the sealed trait approach. I'll think a bit more about to do encode these constant Categorical distributions, and make a PR at some point soon.

PR #393, opened 2019-10-07
Add WartRemover and fix a reasonable set of warts

sritchie:
Please feel free to ignore if you don't find this helpful, or we can back out the set of changes that you do like.

This PR adds support for WartRemover, a nice scala linter: https://www.wartremover.org/

My motivation here was to make sure that all public methods and vals had declared return types, as these are very helpful in understanding Rainier, and in checking binary compatibility with mima once/if Rainier starts doing that.

The changes I completed here were:

- adding return types
- removing implicit toString conversions, in favor if the `s""` macro style
- adding `extends Product with Serializable` to sealed traits - this makes the error messages much nicer, as the compiler can actually infer the particular subclass of the sealed trait, vs going all the way back to `Product with Serialiable` (see here for a nice discussion and rationale: https://nrinaudo.github.io/scala-best-practices/adts/product_with_serializable.html)
- I did have to add ONE trait, `EncoderFrom`, since once I added a proper return type to `Predictor.apply`, the `Encoder` return type prevented anyone from calling `.fit` on the result, specifically this from RegressionBenchmark:

```scala
def build(): RandomVariable[Real] =
    for {
      betas <- RandomVariable.fill(k)(Normal(0, 1).param)
      sigma <- Uniform(0, 10).param
      _ <- Predictor[Double]
        .fromVector(k) { vec =>
          val mean = Real.dot(betas, vec)
          Normal(mean, sigma)
        }
        .fit(data)
    } yield sigma
```


build.sbt
```diff
@@ -9,6 +11,19 @@ lazy val root = project.
 
 scalafmtOnCompile in ThisBuild := true
 
+// see https://www.wartremover.org/doc/warts.html for the full list.
+wartremoverErrors in (Compile, compile) ++= Seq(
```
sritchie, line 15: This isn't the full set of safe warts they recommend checking for, but many of the other ones were too intense to do as a small exercise. It's easy to add more to this list if we want to lock it down even tighter.

rainier-core/src/main/scala/com/stripe/rainier/ir/MethodGenerator.scala
```diff
@@ -96,7 +96,7 @@ private trait MethodGenerator {
   def classNameForMethod(classPrefix: String, id: Int): String = {
     val n = id / classSizeLimit
     if (n > 0)
-      classPrefix + "$" + n
+      s"classPrefix$$$n"
```
sritchie, line 99: a double dollar sign encodes a dollar sign.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
```diff
@@ -93,8 +93,14 @@ object Predictor {
     def fromVector[L](k: Int)(fn: IndexedSeq[U] => L): Predictor[Seq[X], L]
   }
 
-  def apply[X](implicit enc: Encoder[X]) =
-    new From[X, enc.U] {
+  trait EncoderFrom[X, U] extends From[X, U] {
```
sritchie: It MIGHT just be the case that the return types of `From`'s methods should be `EncoderPredictor` already... is that the case?

sritchie:
@avibryant, any thoughts on this one?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

avibryant:
@sritchie sorry for the delay - I'm definitely bullish, but won't have time to review until monday or tuesday.

sritchie:
@avi-stripe, just wanted to ping - would you mind giving this a look? I think there's only one real type question, otherwise this should be pretty routine.

Cheers!

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
avi-stripe: Yes, that is in fact the case, and that would be a better change here.

avi-stripe (changes requested):
Sorry for forgetting about this! It looks good apart from the EncoderPredictor stuff.

rainier-core/src/main/scala/com/stripe/rainier/core/Predictor.scala
sritchie: Okay, excellent, done.

sritchie:
okay, @avi-stripe , this should be good to go.

avi-stripe (approved):

PR #394, opened 2019-10-08
Add support for explicit ordering in several repl methods

andrew-stripe:
This implements versions of `precis`, `coeftab` and `coef` that accept samples as `Seq[(String, Double)]` so that the caller can explicitly order the labels. This makes the output much easier to interpret since it can be in a sensible order.

To make sampling work, I added a generator for strings that mirrors what we do with map keys.


avi-stripe:
I think we could avoid duplicate definitions by using `Traversable[(String,Double)]` rather than either `Seq` or `Map`

avi-stripe (approved):

PR #396, opened 2019-10-09
add Sample and trace

avi-stripe:
This introduces a `Sample` case class which encapsulates the raw parameter samples from a `RandomVariable`, and can either produce the predictive samples from them, or diagnostics. There are new `toSample` methods on RV, and `waic` has been moved to `Sample`. In the future more will be moved to the `Sample` class.

This also adds a `trace` method for use with Jupyter which takes a `Sample` and produces plots like this:

Screen Shot 2019-10-09 at 8 34 04 AM

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
@@ -58,6 +58,27 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
     sampledParams.map(fn)
   }
 
+  def toSample()(implicit rng: RNG): Sample[T] =
+    toSample(Sampler.Default.sampler,
+             Sampler.Default.warmupIterations,
+             Sampler.Default.iterations)
+
+  def toSample(sampler: Sampler,
+               warmupIterations: Int,
+               iterations: Int,
+               keepEvery: Int = 1,
+               nChains: Int = 1,
+               parallel: Boolean = false)(implicit rng: RNG): Sample[T] = {
+    val range = if (parallel) 1.to(nChains).par else 1.to(nChains)
+    val chains =
+      range.map { _ =>
+        Sampler
+          .sample(density(), sampler, warmupIterations, iterations, keepEvery)
+      }.toList
+
+    Sample(chains, targets, targetGroup.variables, value)
+  }
+
   def sample[V]()(implicit rng: RNG, tg: ToGenerator[T, V]): List[V] =
     sample(Sampler.Default.iterations)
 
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Sample.scala
```diff
@@ -0,0 +1,29 @@
+package com.stripe.rainier.core
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.sampler._
+
+case class Sample[+T](chains: List[List[Array[Double]]],
```
andrew-stripe, line 6: Would it be better to favor `Seq` over `List` in these APIs since there's nothing particular about `List` that they need?

rainier-core/src/main/scala/com/stripe/rainier/repl/package.scala
```diff
@@ -194,7 +195,7 @@ package object repl {
       models,
       Sampler.Default.sampler,
       Sampler.Default.warmupIterations,
-      Sampler.Default.iterations,
+      Sampler.Default.iterations
     )
 
   def compare(models: Seq[(String, RandomVariable[_])],
```
andrew-stripe, line 201: I think this form of `compare` should be replaced by one that takes a `Seq[(String, Sample[_])]`. Also, should the models in this form be a vararg to match the other form?

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
```diff
@@ -58,6 +58,27 @@ class RandomVariable[+T](val value: T, val targets: Set[Target]) {
     sampledParams.map(fn)
   }
 
+  def toSample()(implicit rng: RNG): Sample[T] =
```
andrew-stripe, line 61: I think the alternative would be a `Sample.apply(rv, ...)`. Not sure there's any reason to prefer one over the other.

andrew-stripe (approved):
Some minor comments but no need to re-r.

Maybe worth moving `record` over as part of this too?

rainier-core/src/main/scala/com/stripe/rainier/core/Sample.scala
avi-stripe: I think at Stripe we tend to discourage `Seq` in favor of concrete types in most situations.

rainier-core/src/main/scala/com/stripe/rainier/core/RandomVariable.scala
avi-stripe: Yeah, I considered that form, I think this is more in keeping with the API as it exists currently.

rainier-core/src/main/scala/com/stripe/rainier/repl/package.scala
avi-stripe: I agree that we should have a form of `compare` that takes `Sample` but I think I'd like to defer this until a more comprehensive refactoring.

avi-stripe:
@andrew-stripe thank you for the review and comments even if I ended up not taking any of them :)

PR #400, opened 2019-10-23
Fix WAIC bug with identical targets

avi-stripe:
Before this fix, these two equivalent models would have different WAIC:

```
val x12 = List(1.0, 2.0)
val a = Events.observe(x12, x12){x => Normal(x, 1)}
val b = Events.observe(x12, x12, Fn.encode[Double].map{x => Normal(x,1)})
compare("a" -> a, "b" -> b)
```

The bug was that `Sample` was mapping over a `Set[Target]` to get a `WAIC` for each of them, resulting in a `Set[WAIC]` which would only retain one instance of any groups of identical `WAIC` values. This fix forces a change to `List[Target]` within `Sample`, avoiding the issue.


andrew-stripe (approved):
Wow, nice find. I want to suggest adding a comment about why we need the `toList` but I'm struggling to think of a concise way to justify it. If you can, I'd add a comment, otherwise seems good as-is.

PR #402, opened 2019-10-30
Parameter

avi-stripe:
This is a follow up to https://github.com/stripe/rainier/pull/401 .

It introduces a new `Parameter` private subtype of `Variable`, which holds onto its own `density` function. These must be constructed via `Real.parameter{x => density(x)}`.

The intent here is to experiment with alternative APIs that don't need a `RandomVariable` wrapper.

This PR also adds `parameter` methods to `Continuous` which are equivalent to `param` but make use of `Parameter` to return a bare `Real` instead of a `RandomVariable[Real]` (however there's no way yet to sample a model built from this).


andrew-stripe:
This looks good but per Slack should probably land into an experimental branch for now.

PR #405, opened 2019-11-07
Generator optimizations

avi-stripe:
A handful of optimizations that speed up generation of, especially, discrete distributions.

* cache the double form of constant `Real` objects because `BigDecimal.toDouble` is costly
* when generating `Poisson` for large values of `lambda`, use a method that is not O(lambda)
* lazily create the approximate versions of the `Binomial` generator in case we don't need them


rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
```diff
@@ -117,7 +117,10 @@ object Real {
   val negInfinity: Real = NegInfinity
 }
 
-final private[rainier] case class Constant(value: BigDecimal) extends Real
+final private[rainier] case class Constant(value: BigDecimal) extends Real {
+  val toDouble = value.toDouble
```
andrew-stripe: Is there any reasonable case where this is never evaluated such that we're better off making it lazy?

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
```diff
@@ -121,32 +122,60 @@ final case class Poisson(lambda: Real) extends Discrete {
...
+      }
     }
 
-  def logDensity(v: Real) =
-    lambda.log * v - lambda - Combinatorics.factorial(v)
+    0
+  }
+
+  //from https://www.johndcook.com/blog/2010/08/16/how-to-compute-log-factorial/
+  private def logFactorial(n: Int): Double = {
```
andrew-stripe, line 177: This form from the post has an error of `1/12(n+1)`, so at most ~0.04. Is that accurate enough for us? The post suggests using precomputed values for small numbers and then using this approximation (subtracting `1/(12 x)`) for large values.

andrew-stripe (approved):
Looks good assuming my two comments don't lead to anything.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
andrew-stripe: If we don't care about the extra accuracy, probably worth commenting about why

rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
avi-stripe: Very good point. That is a common case and I'll make it lazy.

rainier-core/src/main/scala/com/stripe/rainier/core/Discrete.scala
avi-stripe: Ok, I'll leave a comment. Empirically, there's a very small chance (around 0.1%) that `n` here will ever be < 10, even for a lambda=30 which is the lowest it can be, and that chance would go down for larger lambdas. So I think the bias here would be very slight.

PR #407, opened 2019-12-03
Introduce Model

avibryant:
This PR introduces a `Model` object, with creation methods that parallel the old `Events.observe`. To support those methods, we also reintroduce the logic that used to be in `ToLikelihood` typeclasses, but simplified to be directly on `Distribution`; this makes it more straightforward to add more Distribution subtypes.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
@@ -3,29 +3,33 @@ package com.stripe.rainier.core
 import com.stripe.rainier.compute._
 
 case class Model(targets: List[Target]) {
-    def merge(other: Model) = Model(targets ++ other.targets)
+  def merge(other: Model) = Model(targets ++ other.targets)
 }
 
 object Model {
-    def observe[Y](ys: Seq[Y], dist: Distribution[Y]): Model =  {
-        val target = dist.target(ys)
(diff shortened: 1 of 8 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
```diff
@@ -0,0 +1,35 @@
+package com.stripe.rainier.core
+
+import com.stripe.rainier.compute._
+
+case class Model(targets: List[Target]) {
```
andrew-stripe: Is it intentional that RV uses `Set[Target]` while this uses `List[Target]`?

Separately, do we ever expect the user to access `targets` directly? I'm wondering if it should be `private[core]`.

andrew-stripe (approved):
API looks good modulo one comment. I don't think I'm deep enough into the Target stuff to really evaluate if it's being used correctly.

rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
avibryant: Great question, and I believe I should change this to `Set`. The case we're trying to catch here is something like model1.merge(model2).merge(model1) - that's much easier to accidentally do using the old RV API but still possible here and worth preventing.

Agreed on the access control.

PR #408, opened 2019-12-03
add Prediction

avibryant:
This is a very quick addition to the Model PR which fleshes out the next step: once you have a `Model` from doing an observation, you can `predict(v)` with it, where `v` is a `Real` or a `Map` of `Real`s or a `Generator`, etc. You then get an object that is the equivalent of an `RVG` - you can map/flatMap it as if it were a Generator, to refine your prediction, but it carries the `Model` state around with it to allow inference later.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
@@ -5,8 +5,8 @@ import com.stripe.rainier.compute._
 case class Model(private val targets: Set[Target]) {
   def merge(other: Model) = Model(targets ++ other.targets)
 
-    def predict[T,U](value: T)(implicit tg: ToGenerator[T,U]): Prediction[U] =
-        Prediction(this, tg(value))
+  def predict[T, U](value: T)(implicit tg: ToGenerator[T, U]): Prediction[U] =
+    Prediction(this, tg(value))
 }
 
 object Model {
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
```diff
@@ -4,6 +4,9 @@ import com.stripe.rainier.compute._
 
 case class Model(targets: List[Target]) {
   def merge(other: Model) = Model(targets ++ other.targets)
+
+  def predict[T, U](value: T)(implicit tg: ToGenerator[T, U]): Prediction[U] =
```
andrew-stripe, line 8: To make sure I udnerstand, if T is a generator, then it'll be possible to predict many values for whatever the generator emits?

avibryant: So,

```
...
Model.observe(ys, dist).predict(g)
```
is equivalent to  something like

```
for {
 ...
 _ <- Events.observe(ys, dist)
} yield g
```

So if `g` is a `Generator[T]`, then we will produce a simulated/predicted `T` for each sample. Does that help?

andrew-stripe: Yep, that's what I figured, thanks!

andrew-stripe (approved):

PR #409, opened 2019-12-03
Model.sample

avibryant:
This closes the loop by giving a `toSample` to `Prediction` which does the actual sampling (but still allows modification to the generating function, after sampling, for quick exploration).

This is the minimal change to be able to actually run new-style models. Once this works and is landed I'll go back and revisit the APIs around `Sampler`, `TargetGroup`, etc.


andrew-stripe (commented):
Ah, I think this is getting clearer. I had been thinking you could sample from the `Model` but I guess you need to convert it to a `Prediction` to do the equivalent of the `yield`/map in the old for-loop based API. Is that right?

If so, I'm a bit suspicious about having a separate `Model` and `Prediction` since the only two things `Model` does is create a `Prediction` and merge with other models (which is a lot like `Prediction#zip`). Maybe there's something around adding a `generator` to `Model`, and having `Model.observe` do something along the lines of setting the generator to `Generator.constant(Unit)`. `zip` gets a little weird because you'll end up with a wacky generator containing nested tuples of `Unit` but maybe that doesn't matter since you'll just ignore them when you go to `map` in there generator you actually want?

avibryant:
Yes, you're asking the right question, and I've gone down the same train of thought myself. I go back and forth on this. My argument in favor of the current approach is:

* first, the whole Unit thing does seem clunky to me
* second, I think it's a bit clearer to separate out the Model stage from any predictions you might want to make with it; rather than returning a `Prediction[T]` where you've already locked in a particular prediction, you could return a `Model` plus some structure that contains some `Real`s for your parameters etc, and leave someone else to build a prediction from it.

Another point in the design space would be to get rid of Prediction and just have `sample(g: Generator[T]): Sample[T]` on `Model`; and if you want to pass around a not-yet-sampled Model + Generator you just do that as a tuple of those two things.

Thoughts?

andrew-stripe:
I think I like the idea of starting with: 
> get rid of Prediction and just have `sample(g: Generator[T]): Sample[T]` on `Model`

At worse, `Prediction` comes back later and this ends up as sugar for the very common case of `model.predict(g).toSample`

avibryant:
@andrew-stripe so this does something even more minimal, which is that `Model` has `sample()` which only takes sampler parameters, no generator, and then returns a `Sample`; and then you can run `predict(g)` on `Sample` to get a list of values.

rainier-core/src/main/scala/com/stripe/rainier/core/Sample.scala
```diff
@@ -1,25 +1,19 @@
...
-import com.stripe.rainier.compute._
 import com.stripe.rainier.sampler._
 
-case class Sample[+T](chains: List[List[Array[Double]]],
-                      targets: List[Target],
-                      variables: List[Variable],
-                      value: T) {
-  def map[U](fn: T => U) = Sample(chains, targets, variables, fn(value))
-
+case class Sample(chains: List[List[Array[Double]]], model: Model, rng: RNG) {
```
andrew-stripe: Should the RNG be implicit?

avibryant: Yeah, might as well be. Will change.

andrew-stripe (approved):
Minor comment but looks good. Don't forget to delete Prediction.scala

PR #410, opened 2019-12-04
Bring back SBC

avibryant:
This is an important piece to be able to bring back benchmarks and tests.

Rather than looking at the full file here, you can use this diff to see what's changed since before it was deleted:
https://github.com/stripe/rainier/compare/f71a8a705da55148a2a10436796c9f90e108f1a9..991f65d3650195d6f910fd8925f8888015f01486#diff-c279ee7bd4b82e7ee0df025715f7fddf

(This is a fairly small and straightforward update to the new API).


andrew-stripe (approved):
Looks like you lost the `posteriorSamples` method. I assume that was intentional?

avibryant:
Yeah I don't think it ever got used so seemed simpler to delete than to update.

PR #420, opened 2019-12-06
Track bounds for Real

avibryant:
This tracks a (lower,upper) bound (as a double) for any `Real`, by starting each `Variable` at (-inf,inf), and starting each `Constant(k)` at (k,k), and then doing interval arithmetic to transform the bounds for any operations.

It works surprisingly well - eg `Exponential(3).param.logistic.bounds` gives `Bounds(0.5,1.0)`.

The idea is then to have assertions in the constructors for distributions that enforce bounds like stddev has to be positive. This acts a bit like type checking for your models.

Before this is useful, we'll need to also add accurate bounds to `Placeholder` that are computed during `Model.observe`, otherwise we'll likely get assertion failures every time we use `Fn`. (We also need to add in the actual assertions).

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
@@ -241,9 +241,10 @@ Evaluates to the (index-low)'th element of table.
  */
 private final class Lookup(val index: NonConstant,
                            val table: Array[Real],
-                           val low: Int,
-                           val bounds: Bounds)
-    extends NonConstant
+                           val low: Int)
+    extends NonConstant  {
+      val bounds = Bounds.or(table.map(_.bounds))
+  }
 
 object Lookup {
   def apply(table: Seq[Real]): Real => Real =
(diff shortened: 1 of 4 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
```diff
@@ -17,6 +16,16 @@ Apart from Variable and a simple ternary If expression, all of the subtypes of R
 You can also automatically derive the gradient of a Real with respect to its variables.
  */
 sealed trait Real {
+  def bounds: Bounds
+  def mustBePositive(): Unit = {
```
andrew-stripe: Maybe these assertions should be on Bounds instead of Real? `myReal.bounds.mustBePositive` feels like a reasonable API and keeps the assertions separate from the compute graph.

Total nit, but should the name be something like `assertPositive` to match `assert` in the language?

andrew-stripe (approved):
Makes sense! Potentially would have avoided quite a few broken models I've written.

I think under certain conditions this might allow you to assert other things like the ordering of two parameters (e.g. `a = b + c` and we know `c` to be positive, then we know `b > c`). 

This looks very amenable to some table-driven tests.

avibryant:
I'll add an issue for adding tests for this.

rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
avibryant: Yeah, maybe what I'll do is just add `isPositive` and `betweenZeroAndOne` to `Bounds`, and then we can use that with `assert` on the caller (and if we want to ramp that down we can introduce a `warn` or something).

PR #426, opened 2019-12-13
be explicit about parameters vs placeholders

avibryant:
This just makes it clear where we're using `Parameter` vs `Placeholder` vs the generic `Variable`.

It also deletes some of the Viz stuff to be reintroduced later with support for placeholders and bounds.

This is setting things up for some larger changes around `Placeholder`.


rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
```diff
@@ -1,14 +1,11 @@
 package com.stripe.rainier.compute
 
 import com.stripe.rainier.ir._
-import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.HashMap
 
 private object Gradient {
 
-  def derive[F[α] <: TraversableOnce[α]](variables: F[Variable], output: Real)(
```
andrew-stripe, line 9: Why drop the CanBuildFrom and whatnot?

andrew-stripe (approved):
It's difficult to review in detail whether this hits all the right references. Are the specific things you'd want me to look at in detail?

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
avibryant: Frankly just because it's stylistically fancier and more general than I think is needed or than is standard in this codebase, and so when I ran across a need to change it, I took the opportunity to get rid of it.

avibryant:
Nothing here that I think needs review in detail; this is more just for your awareness. The next PR will be more interesting.

PR #427, opened 2019-12-14
Placeholder(values), and fold Encoder into Fn

avibryant:
The motivating change here was to attach the `Array[Double]` of values directly to `Placeholder`, so that it could compute its `bounds` correctly.

One consequence of this is that now, in the same way that we discover a `Parameter`'s prior through reachability (ie by traversing the graph from the likelihood), we can discover which data we need to include in the model the same way; this is nice when you might be, eg, encoding all of a large case class, but only actually referencing one or two fields in your model. Currently, to minimize the diff, this happens during the creation of a `Target` (which used to get passed the data explicitly).

There are two other changes rolled into this PR because the Placeholder change required major surgery in their areas and so it seemed more efficient to move things forward while I was at it.

The first and less significant is a couple of optimizations that were dropped from Target and DataFunction. In the short term this makes things significantly slower, but my intent is to bring them back, implemented at a higher layer of the system, which will allow them to be more flexible and have less complexity cost. The two optimizations are the dynamic inlining of likelihood computation, and the `batchBits` partial loop unrolling of the same.

The second is to merge `Encoder` into `Fn`. Using a `Fn` is now the only way to encode data for batch computation, and the mechanics of doing so can be wholly encapsulated inside `Fn` instead of being fairly openly accessible in `Encoder`.

This means that `Distribution[T]` now just needs a `likelihoodFn: Fn[T,Real]` instead of a lot of encoder-related stuff. (This then can generally delegate to a `logDensity`).

For now I've chosen to go with an explicit API for building `Fn`s for complex data types, rather than the implicit API that `Encoder` had. So for example instead of

```scala
Fn.encode[(Int,Int)]
```

You need to do

```scala
Fn.numeric[Int].zip(Fn.numeric[Int])
```

We can certainly bring back more magical implicits later if we want; however, since they can't do everything we'd want (eg, for a `List[Int]` we need to specify the size of the list, which you can't do implicitly), I thought we should try going all the way explicit.


avibryant:
https://github.com/stripe/rainier/pull/427/commits/67b4eeab515e8404fcbfb04ee7893cff4d16b5ed shows how dramatically slower this can be, having removed the optimizations. Future PRs should get things back to the way it was before.

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
```diff
@@ -0,0 +1,94 @@
...
+
+      def xy(x: X) = g(self.xy(x))
+    }
+
+  def keys[K](seq: Seq[K]): Fn[Map[K, A], Map[K, Y]] =
+    new Fn[Map[K, A], Map[K, Y]] {
+      type X = Map[K, self.X]
+      def wrap(a: Map[K, A]) = a.map { case (k, v) => k -> self.wrap(v) }
+      def create(columns: List[Array[Double]]) =
+        ??? //seq.map{k => k -> self.create(columns)}.toMap
```
andrew-stripe: Did you mean to leave this unimplemented?

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
```diff
@@ -0,0 +1,94 @@
...
+      def xy(x: Real) = x
+    }
+
+  def enum[T](choices: List[T]): Fn[T, List[(T, Real)]] =
+    new Fn[T, List[(T, Real)]] {
+      type X = List[(T, Real)]
+      def wrap(a: T) = choices.map { k =>
+        if (a == k) (k, Real.one) else (k, Real.zero)
+      }
+      def create(columns: List[Array[Double]]) = ???
```
andrew-stripe: Did you mean to leave this unimplemented?

rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
```diff
@@ -73,10 +63,11 @@ object Real {
     summed.log + max
   }
 
-  def placeholder(): Variable = new Placeholder()
+  def placeholder(values: Array[Double]): Placeholder = new Placeholder(values)
```
andrew-stripe, line 66: Does this more tightly tie the Rainier API to having everything in-memory than it already is? Maybe it's fine because we can allow a version that just needs a min/max in the future?

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
```diff
@@ -0,0 +1,94 @@
...
+  protected type X
+
+  protected def wrap(a: A): X
+  protected def extract(a: A, acc: List[Double]): List[Double]
+  protected def create(columns: List[Array[Double]]): (X, List[Array[Double]])
+  protected def xy(x: X): Y
+
+  def apply(a: A): Y = xy(wrap(a))
+
+  def encode(as: Seq[A]): Y = {
```
andrew-stripe: I'm not sure as `encode` is a good descriptive name for what this does.

andrew-stripe (approved):
Possibly minor comments but happy to re-r if they lead to bigish changes.

I'm also thinking about what the interface will look like for creating a custom encoder for your own case class. I think it's something like:
```
case class MyThinger(...) {
}

object MyThinger {
  val fn = new Fn[MyThinger, Real] {
    ...
  }
}

case class MyThingerEnc(...)
```

Assuming that's roughly right it seems no worse (and maybe better?) than the status quo.

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
avibryant: No, thanks, willfix.

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
avibryant: No, thanks, willfix2.

rainier-core/src/main/scala/com/stripe/rainier/compute/Real.scala
avibryant: Yeah, I think in the future we'd just introduce a new subtype of `Variable`.

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
avibryant: Fair yep. Any suggestions?

andrew-stripe: I was afraid you'd ask that.

One other thing that seems odd is that both `apply` and `encode` return a `Y`. Maybe it makes sense to get rid of `apply` and rename `encode` to `apply`? I assume Seq will be the more common case and it's easy to `fn(Seq(v))` if necessary.

avibryant: We still need `apply` as a public API, in particular if you want to use the `Fn` as part of prediction, but just in general if you want to treat it as a normal function. Whereas `encode` is closer to a private API (and in fact we could maybe make it `private[rainier]`), that has to be used with care and I do not expect end users to make use of it.

andrew-stripe: Ah OK, I got my usage mixed up then. I'm probably less concerned about the naming if it can be `private[rainier]`. I think I've lost track of what `encode` is doing here.

Relatedly, do we need `wrap` at all or can we `def apply(a: A): Y = encode(Seq(a))`?

avibryant: @andrew-stripe an example will help here.

Let's take the simple case of `f = Fn.numeric[Double].map{x => x + 1}`.

The result of `f.wrap(1.0)` is `Constant(1.0)`.

The result of `f.apply(1.0)` will be `Constant(2.0)`.

The result of `f.encode(List(1.0))` is a Real that is equivalent to `Placeholder(List(1.0)) + 1`.

Does that help?

avibryant:
Finished those Fn implementations as well as adding tests which makes me much more confident in them.

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
andrew-stripe: Ah yea, I was getting tripped up on the fact that the return type is the same in both cases. In the example you described it's always `Real` but those represent different compute graphs between `wrap` and `encode`.

avibryant:
The tests are timing out, presumably because of the lack of optimizations (it only takes 30s to run on my MBP but the container travis is using must have a lot less CPU). I'll try to bring some of them back to get things green again.

PR #431, opened 2019-12-17
wrap BigDecimal in Decimal

avibryant:
This is a first step towards eliminating BigDecimal: it wraps all uses of BigDecimal in our own Decimal object.

### DIFF
--- a/rainier-benchmark/src/main/scala/com/stripe/rainier/bench/CategoricalBenchmark.scala
+++ b/rainier-benchmark/src/main/scala/com/stripe/rainier/bench/CategoricalBenchmark.scala
@@ -1,0 +1,31 @@
+package com.stripe.rainier.bench
+
+import org.openjdk.jmh.annotations._
+import java.util.concurrent.TimeUnit
+
+import com.stripe.rainier.compute._
+import com.stripe.rainier.core._
+import com.stripe.rainier.sampler._
+
+@BenchmarkMode(Array(Mode.SampleTime))
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 3, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(1)
+@Threads(4)
+@State(Scope.Benchmark)
+class CategoricalBenchmark {
+  implicit val rng: RNG = RNG.default
+  implicit val eval: Numeric[Real] = new Evaluator(Map.empty)
+
+  @Benchmark
+  def run(): Unit = {
+    Categorical
+      .list(1.to(100))
+      .generator
+      .flatMap { i =>
+        Categorical.list(1.to(i)).generator
+      }
+      .get
+  }
+}
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
@@ -1,0 +1,47 @@
+package com.stripe.rainier.compute
+
+case class Decimal(value: BigDecimal) {
+    def toDouble = value.toDouble
+    def toInt = value.toInt
+
+    def isValidInt = value.isValidInt
+    def isWhole = value.isWhole
+
+    def abs = Decimal(value.abs)
+
+    def pow(exponent: Int) =
+        Decimal(value.pow(exponent))
+        
+    def >(other: Decimal) =
+        value > other.value
+    def <(other: Decimal) =
+        value < other.value
+    def <=(other: Decimal) =
+        value <= other.value
+    def >=(other: Decimal) =
+        value >= other.value
+
+    def +(other: Decimal) =
+        Decimal(value + other.value)
+
+    def -(other: Decimal) =
+        Decimal(value - other.value)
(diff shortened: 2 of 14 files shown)


avibryant:
cc @sritchie

avibryant:
It's failing on compiling rainier-cats right now, I'll fix that in a bit

rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
```diff
@@ -178,6 +178,10 @@ class RealTest extends FunSuite {
       (x * x * x))
   }
 
+  run("x^(10(2/10 + 1/10))") { x =>
```
avibryant, line 181: This currently passes, but would fail with a `Decimal(value: Double)` implementation.

rainier-tests/src/test/scala/com/stripe/rainier/compute/RealTest.scala
```diff
@@ -181,7 +181,11 @@ class RealTest extends FunSuite {
   run("x^(10(2/10 + 1/10))") { x =>
     x.pow((Real.one / 10 + Real.two / 10) * 10)
   }
-
+/*
+  run("x^((4/3 + 1)*3)") { x =>
```
avibryant, line 185: This fails even with BigDecimal, but we might be able to get it to pass with a different `Decimal` implementation

rainier-benchmark/src/main/scala/com/stripe/rainier/bench/CategoricalBenchmark.scala
```diff
@@ -0,0 +1,31 @@
...
+@Measurement(iterations = 5, time = 1)
+@Fork(1)
+@Threads(4)
+@State(Scope.Benchmark)
+class CategoricalBenchmark {
+  implicit val rng: RNG = RNG.default
+  implicit val eval: Numeric[Real] = new Evaluator(Map.empty)
+
+  @Benchmark
+  def run(): Unit = {
```
avibryant, line 22: Switching Decimal to use `Double` instead of `BigDecimal` is a 6x speedup on this.

sritchie: Fantastic! I'll convert the example I've been using as a benchmark into an actual benchmark, but it looks like this will be plenty fast once the changes settle.  Once I review these I can link you to what I've doing; I'm excited to use Rainier to implement a Bayesian bandit.

rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
```diff
@@ -0,0 +1,47 @@
+package com.stripe.rainier.compute
+
+case class Decimal(value: BigDecimal) {
```
sritchie, line 3: You should be able to change this to:

```
case class Decimal(value: BigDecimal) extends AnyVal { ... }
```

to make this into a value class and avoid any extra allocations at all.

rainier-core/src/main/scala/com/stripe/rainier/compute/Gradient.scala
```diff
@@ -141,7 +141,7 @@ private object Gradient {
           LogLine(complement)
       gradient.toReal *
         exponent *
-        term.pow(exponent - 1) *
+        term.pow(exponent - Decimal(1)) *
```
sritchie: `Decimal.One`, use it since you've got it! It won't matter if you make the `AnyVal` change, since you won't be doing any allocation... but looks nice anyway.

rainier-core/src/main/scala/com/stripe/rainier/compute/LineOps.scala
```diff
@@ -68,7 +68,7 @@ private[compute] object LineOps {
   def log(line: Line): Option[Real] =
     line.ax match {
       case Coefficients.One(x, a)
-          if (a >= Real.BigZero) && (line.b == Real.BigZero) =>
+          if (a >= Decimal.Zero) && (line.b == Decimal.Zero) =>
         Some(x.log + Math.log(a.toDouble))
```
sritchie, line 72: Can you do `a.log` here? Does that push the conversion and `Math.log` inside the `+` implementation in `Real`?

sritchie (approved):
Looks great! Couple small comments... the `AnyVal` comment is the main one.

rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
avibryant: Yeah, I didn't bother with that since this is just a temporary stage anyway (and https://github.com/stripe/rainier/pull/432 won't be able to do that)

rainier-core/src/main/scala/com/stripe/rainier/compute/LineOps.scala
avibryant: `a` is a `Decimal` (previously a `BigDecimal`) which doesn't have `log`

PR #432, opened 2019-12-18
Decimal implementation with double and fraction reprs

avibryant:
@sritchie I'd love your eyes on this one

This is a follow up to https://github.com/stripe/rainier/pull/431 . For context, this would not be merged into develop, but into a 0.3-dev branch.

It introduces two representations for constants: one is just a double, and the other is a numerator/denominator which are both long. If you create a Real with an integer or long, it will stay in the fraction form as long as you continue to do operations on it with other ints/longs. As soon as you introduce a double, it will convert into double and stay there (we never check to see if we can convert the other way).

This should catch the most common cases where we otherwise lose precision through floating point errors, without hampering performance of the common case of doubles. From simple benchmarking, this is not appreciably slower than just using doubles directly in Real.

Note: I disabled the CategoricalSuite in cats because it was failing on a -Inf != -Inf, and I wasn't sure how best to address that.

### DIFF
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
@@ -3,6 +3,11 @@ package com.stripe.rainier.compute
 import scala.annotation.tailrec
 
 sealed trait Decimal {
+  def ==(other: Decimal) =
+    toDouble == other.toDouble
+  override def hashCode(): Int =
+    toDouble.hashCode()
+
   def toDouble: Double
   def toInt: Int = toDouble.toInt
 
--- a/rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
+++ b/rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
@@ -1,47 +1,121 @@
 package com.stripe.rainier.compute
 
-case class Decimal(value: BigDecimal) {
-  def toDouble = value.toDouble
-  def toInt = value.toInt
+import scala.annotation.tailrec
 
-  def isValidInt = value.isValidInt
-  def isWhole = value.isWhole
+sealed trait Decimal {
+  def toDouble: Double
+  def toInt: Int = toDouble.toInt
 
-  def abs = Decimal(value.abs)
-
-  def pow(exponent: Int) =
-    Decimal(value.pow(exponent))
+  def isValidInt = toDouble.isValidInt
+  def isWhole = toDouble.isWhole
 
   def >(other: Decimal) =
-    value > other.value
+    toDouble > other.toDouble
   def <(other: Decimal) =
-    value < other.value
+    toDouble < other.toDouble
   def <=(other: Decimal) =
-    value <= other.value
+    toDouble <= other.toDouble
   def >=(other: Decimal) =
-    value >= other.value
+    toDouble >= other.toDouble
 
-  def +(other: Decimal) =
-    Decimal(value + other.value)
+  def abs: Decimal =
+    Decimal.abs(this)
+  def pow(exponent: Int): Decimal =
+    Decimal.pow(this, exponent)
 
-  def -(other: Decimal) =
-    Decimal(value - other.value)
+  def +(other: Decimal): Decimal =
+    Decimal.add(this, other)
 
-  def *(other: Decimal) =
-    Decimal(value * other.value)
+  def -(other: Decimal): Decimal =
+    Decimal.subtract(this, other)
 
-  def /(other: Decimal) =
-    Decimal(value / other.value)
+  def *(other: Decimal): Decimal =
+    Decimal.multiply(this, other)
+
+  def /(other: Decimal): Decimal =
+    Decimal.divide(this, other)
+}
+
+case class DoubleDecimal(toDouble: Double) extends Decimal
+case class FractionDecimal(n: Long, d: Long) extends Decimal {
+  lazy val toDouble = n.toDouble / d.toDouble
 }
 
 object Decimal {
-  def apply(value: Double): Decimal =
-    Decimal(BigDecimal(value))
-  def apply(value: Int): Decimal =
-    Decimal(value.toDouble)
+  def apply(value: Double): Decimal = DoubleDecimal(value)
+  def apply(value: Int): Decimal = FractionDecimal(value.toLong, 1L)
+  def apply(value: Long): Decimal = FractionDecimal(value, 1L)
+
+  def abs(x: Decimal): Decimal = x match {
+    case DoubleDecimal(v)      => DoubleDecimal(Math.abs(v))
+    case FractionDecimal(n, d) => FractionDecimal(n.abs, d.abs)
+  }
+
+  def pow(x: Decimal, y: Int): Decimal = x match {
+    case DoubleDecimal(v) => DoubleDecimal(Math.pow(v, y.toDouble))
+    case FractionDecimal(n, d) =>
+      FractionDecimal(
+        Math.pow(n.toDouble, y.toDouble).toLong,
+        Math.pow(d.toDouble, y.toDouble).toLong
+      )
+  }
+
+  def add(x: Decimal, y: Decimal): Decimal = (x, y) match {
+    case (DoubleDecimal(v), _) => DoubleDecimal(v + y.toDouble)
+    case (_, DoubleDecimal(v)) => DoubleDecimal(v + x.toDouble)
+    case (FractionDecimal(n1, d1), FractionDecimal(n2, d2)) => {
+      val d = lcm(d1, d2)
+      val n = (n1 * d / d1) + (n2 * d / d2)
+      FractionDecimal(n, d)
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
```diff
@@ -1,47 +1,129 @@
 package com.stripe.rainier.compute
 
-case class Decimal(value: BigDecimal) {
-  def toDouble = value.toDouble
-  def toInt = value.toInt
+import scala.annotation.tailrec
 
-  def isValidInt = value.isValidInt
-  def isWhole = value.isWhole
+sealed trait Decimal {
```
sritchie, line 5: odd, I would have thought that the `WartRemover` would make this comment - it should be running on compile.

If you use 

```
sealed trait Decimal extends Product with Serializable
```

type errors won't infer "Product with Serializable" as the type of case classes that extend this trait, and errors will be more nicely readable.

(Maybe that only applies to traits with generics?)

rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
```diff
@@ -1,47 +1,129 @@
...
+  def ==(other: Decimal) =
```
sritchie, line 6: what do you think of return types on the public API?

rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
```diff
@@ -1,47 +1,129 @@
...
-  def /(other: Decimal) =
-    Decimal(value / other.value)
+  def *(other: Decimal): Decimal =
+    Decimal.multiply(this, other)
+
+  def /(other: Decimal): Decimal =
+    Decimal.divide(this, other)
+}
+
+case class DoubleDecimal(toDouble: Double) extends Decimal
```
sritchie, line 44: you know... I don't have great intuition here for how much this would help, but I THINK if you make `Decimal` extend `Any`:

```scala
sealed trait Decimal extends Any with Serializable
```

then this can be a value class:

```scala
case class DoubleDecimal(toDouble: Double) extends AnyVal with Decimal
```

Here are the docs: https://docs.scala-lang.org/overviews/core/value-classes.html

I THINK this will avoid allocation unless you call one of the methods, which we almost certainly will do. I can't understand what the answer to this question is suggesting, but it seems like it's possible to avoid that penalty: https://stackoverflow.com/questions/47140707/value-classes-universal-traits-and-the-necessity-of-instantiation?rq=1

rainier-core/src/main/scala/com/stripe/rainier/compute/ToReal.scala
```diff
@@ -33,6 +33,16 @@ trait LowPriToReal {
 object ToReal extends LowPriToReal {
   def apply[A](a: A)(implicit toReal: ToReal[A]): Real = toReal(a)
 
+  implicit val fromInt: ToReal[Int] =
```
sritchie, line 36: Should we add a `fromDouble` as well?

sritchie (approved):
This is great!

sritchie:
@avibryant, there are two main typeclasses I use with a categorical distribution that might belong in rainier - `Expectation` and `Decompose`. I'm curious if there is a more natural way to accomplish this, so let me describe the two traits.

Two main things make sense to model, sometimes, as categorical distributions:

- the policy an agent takes given some state of the world is a distribution over actions the agent could take, `Categorical[A]`
- some worlds can model their state transitions as `Categorical[(Reward, NextState)]`, and can associate one of those instances with each object.

### Expectation
I use this `Expectation` typeclass to generate expected values, given some `Categorical`. It's generic so it can work for `cats.Id[A]` instances, my categorical distribution, rainier's, or some new thing that comes along.

- `f` is a function from some outcome to its outcome value... so I guess that's the "random variable"
- I use a `Module[Double, B]` to scale each outcome value by its probability, then sum them all up to get my expectation.

I could back `Double` out to `Decimal` or `Real` now.

### Decompose
I added a [`Decompose`](https://github.com/sritchie/scala-rl/blob/develop/scala-rl-core/src/main/scala/com/scalarl/algebra/Decompose.scala) typeclass to break some `M[A]` down into a linear combination of `(A, R)` pairs... experiment outcome and probability. That feels like an idea that comes up a lot.

```scala
trait Decompose[M[_], R] extends Serializable {
  def ring: Ring[R]
  def decompose[A](ma: M[A]): Iterator[(A, R)]
}
```

What do you think? Is there a better way to get an expected value out of a categorical distribution? My stats vocab is not great (yet!) so it might be staring me in the face.

rainier-core/src/main/scala/com/stripe/rainier/compute/Decimal.scala
avibryant: I am in favor of them :) Missed this one, willfix.

rainier-core/src/main/scala/com/stripe/rainier/compute/ToReal.scala
avibryant: That's covered by the `numeric` in the lowpri. I needed to add int/double here so that they create FractionDecimal instances.

sritchie:
@avibryant, is this why those tests are failing? Do we now have a `Constant(Double.NEG_INF)` or whatever that is in not-pseudocode?

```scala
  def eqReal(epsilon: Double): Eq[Real] = {
    val bde = eqDecimal(epsilon)
    Eq.instance { (left, right) =>
      (left, right) match {
        case (Infinity, Infinity) | (NegInfinity, NegInfinity) => true
        case (Constant(a), Constant(b))                        => bde.eqv(a, b)
        // TODO - the Eq instance returned here doesn't test for full
        // DAG equality; this final case needs to be expanded to
        // compare the various NonConstant options.
        case (a, b) => a == b
      }
    }
  }
```

https://github.com/stripe/rainier/blob/73dee58cd36c0229611f8b12bb90638061ae9abc/rainier-cats/src/main/scala/com/stripe/rainier/cats/package.scala#L125

avibryant:
@sritchie yes, but I don't think we necessarily want to paper over it by explicitly checking for infinities. Previously we introduced the Infinity and NegInfinity subtypes of Real because BigDecimal cannot represent infinities, and tried to make it impossible to ever end up with Constant(inf). Somehow the change to Decimal is making that possible again here, which we may want to treat as a bug.

sritchie:
@avibryant yup, looks like the Eq instance giving us trouble here too https://travis-ci.org/stripe/rainier/jobs/626809074#L362

avibryant:
@sritchie re `Expectation` and `Decompose`, the regime Rainier is designed for is one where you cannot compute exact probabilities for outcomes, but instead can only generate samples - so expectations would be estimated by averaging over the sampled outcomes, rather than ever having probabilities explicitly attached. Categorical with constant probabilities (ie where the probabilities are not themselves continuous random variables) is an extreme special case where you have discrete outcomes with known probabilities and so can actually compute probabilities exactly; but that case is the least interesting one to rainier and it wouldn't make sense to have code for it.

sritchie:
Yeah, that makes sense. This came up in the more boring contrived bits of tabular RL, where you keep cycling through the sequence of all states and updating your estimates based on expectations of what reward and estimated next state value each action might get you.

A more interesting case that I do need to cover next is a case where you have:

- some discrete set of actions you can take, and some weight associated with each (the policy's chance of taking that action, say)

And you can only take a single action. You take the action, and eventually you get some return... then you update your estimate if the value of the action by

- weighting that return by the probability you would have taken it,
- averaging that with your stored estimate of the value of each action you didn't take, weighted by the probability associated with each.
- updating the probabilities based on the new value estimates

that is a more interesting use case. Is that only useful as a conceptual model for something that exists in Rainier? We're sampling and updating estimates based on the same, but we still have this problem of aggregating using the probabilities as weights.

I'm not really sure what I'm asking... I can write code that will work, BUT I'd rather shift my mental model and reuse existing concepts. I'd love any pointers!

avibryant:
@sritchie there's a whole bunch of interesting stuff here to talk about, though we should probably move it to a different medium. The short version, though, is that what you're describing - both because it involves mostly discrete distributions, and because it's about modeling an open-ended process that is probably intricately tied into a simulation - is a really poor fit for Rainier's design choices (Rainier is largely built around the constraints of Hamiltonian Monte Carlo, which is for continuous parameters, fixed model structure, and seeing all the data at once - basically the polar opposite of what you're doing).

In the probabilistic programming world, the techniques that are better suited to this are particle filters, aka Sequential Monte Carlo. There's lot of systems out there geared toward that - in fact, the first version of Rainier was, based on porting this paper to Scala: http://mlg.eng.cam.ac.uk/pub/pdf/SciGhaGor15.pdf Another one that looks really cool is https://birch-lang.org/ (which is its own language), and there's https://probprog.github.io/anglican/ for Clojure.

avibryant:
@sritchie I tried hard to get a relaxed enough `Eq[Decimal]` to let the `Categorical` suite run, but whatever is generating random examples is doing too good a job of exposing floating point numerical instability and it's just not tenable to run the tests.

PR #433, opened 2019-12-18
WIP: do not distribute over multiplications automatically

avibryant:
Again on the theme of hoisting optimizations up into higher layers: previously, if we had something like `(x+1)(x+2) + (y+1)(y+2)` we would distribute out the multiplications before doing the sum. In the case that `x=y`, this will simplify nicely, but otherwise it just makes things worse.

Now, we never do this automatically, but added a `distribute` method to `Real` so you can request it if you know it will be a good idea.


avibryant:
Marking this WIP because I'm realizing that it should distribute recursively instead of just the top-level LogLine.

andrew-stripe:
Should I hold off reviewing for now then?

avibryant:
@andrew-stripe yes, thanks - it's going to get hairier, might as well wait until then.

avibryant:
closing this for now, will revisit later

PR #434, opened 2019-12-20
Inline based on static analysis

avibryant:
The major new thing here is `RealOps.inlinable` which walks the DAG to determine if there are any sub-graphs which a) contain at least one parameter b) contain at least one placeholder and c) have as their root node a non-linear operation like log, exp, pow, etc.

If there are *not* any such sub-graphs, then it should be possible to simplify the sum across all N placeholder rows to a single O(1) function of the parameters. If any such subgraphs do exist, then we need to keep it in placeholder form and do the N independent evaluations.

This PR also tweaks PartialEvaluator to be specifically for inlining placeholder values (and removes that recently-introduced capability from Evaluator).


rainier-benchmark/benchmarks.txt
```diff
@@ -1,44 +1,45 @@
...
-run                  LogNormal                 100000                     30000.000
-run                  NegativeBinomial          100                           20.000
-run                  NegativeBinomial          1000                         200.000
-run                  NegativeBinomial          10000                       2000.000
-run                  NegativeBinomial          100000                     20000.000
-run                  Normal                    100                           20.000
-run                  Normal                    1000                         300.000
-run                  Normal                    10000                       3000.000
-run                  Normal                    100000                     30000.000
+run                  Bernoulli                 100                            0.300
```
avibryant: This now looks like it did before the optimization-removal in https://github.com/stripe/rainier/pull/427

rainier-core/src/main/scala/com/stripe/rainier/compute/Fn.scala
```diff
@@ -12,7 +12,7 @@ trait Fn[-A, +Y] { self =>
 
   def apply(a: A): Y = xy(wrap(a))
 
-  private[rainier] def encode(as: Seq[A]): Y = {
+  def encode(as: Seq[A]): Y = {
```
avibryant, line 15: removing the `private` because it's too useful from the REPL

rainier-tests/src/main/scala/com/stripe/rainier/core/SBCModel.scala
```diff
@@ -53,16 +53,16 @@ object SBCLogNormal extends SBCModel[Double] {
...
-      0.12809334825262858, 0.1302285851887811, 0.12971488966061975,
-      0.12858030863741665, 0.12974670510210187, 0.12864963330710052,
-      0.13046537194900923, 0.12880773413827598, 0.12985465134247504,
-      0.12844225339631635, 0.1313974020144121, 0.12744884255033428,
-      0.1311196525734243, 0.12853622813696675, 0.12853622813696675,
-      0.12977895561929287, 0.12907650766771345, 0.12983463336804282,
-      0.12873418321929223, 0.12873418321929223, 0.13013268941844386,
-      0.12923546640482494, 0.12958663528945488, 0.12958663528945488,
-      0.1300053138358873, 0.12979024604560313, 0.12983475193607927)
+    List(0.7757238193473195, 0.7757238193473195, 0.7757238193473195,
```
avibryant: I'm really confused as to what made the goldset change here, but I'm satisfied that this one is correct.

rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
```diff
@@ -244,4 +244,76 @@ private[compute] object RealOps {
...
     vars.toSet
   }
+
+  //see whether we can reduce this from a function on a matrix of
+  //placeholder data (O(N) to compute, where N is the rows in the matrix)
+  //to an O(1) function just on the parameters; this should be possible if the function can be expressed
+  //as a linear combination of terms that are each functions of either a placeholder,
+  //or a parameter, but not both.
+  def inlinable(real: Real): Boolean = {
+    var seen = Map.empty[Real, (Boolean, Boolean, Boolean)]
```
andrew-stripe: Maybe replace the 3-tuple with a case class for clarity? The case class could also support a `||` operation that applies across all its elements.

rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
```diff
@@ -244,4 +244,76 @@ private[compute] object RealOps {
...
+      rs.map(loop).reduce { (a, b) =>
+        (a._1 || b._1, a._2 || b._2, a._3 || b._3)
+      }
+
+    def loopNonLinear(rs: List[Real]): (Boolean, Boolean, Boolean) = {
+      val (pr, ph, nl) = rs.map(loop).reduce { (a, b) =>
+        (a._1 || b._1, a._2 || b._2, a._3 || b._3)
+      }
+
+      if (pr && ph)
```
andrew-stripe: I'm not sure if this would be clearer written as `(pr, ph, nl || (pr && ph))`

rainier-benchmark/src/main/scala/com/stripe/rainier/bench/misc/CategoricalBenchmark.scala
```diff
@@ -1,4 +1,4 @@
-package com.stripe.rainier.bench
+package com.stripe.rainier.bench.misc
```
andrew-stripe, line 1: Always a little skeptical of things called "misc" becoming dumping grounds, but I don't think it really matters here.

rainier-core/src/main/scala/com/stripe/rainier/core/Sample.scala
```diff
@@ -6,13 +6,6 @@ case class Sample(chains: List[List[Array[Double]]], model: Model)(
     implicit rng: RNG) {
   def diagnostics = Sampler.diagnostics(chains)
 
-  def waic =
```
andrew-stripe, line 9: Dumping waic because it doesn't seem useful in general?

rainier-tests/src/test/scala/com/stripe/rainier/compute/InlineTest.scala
```diff
@@ -0,0 +1,53 @@
...
+  val rng = new scala.util.Random()
+
+  def run(df: DataFunction, params: Array[Double]): Double = {
+    val globals = new Array[Double](df.numGlobals)
+    val outputs = new Array[Double](df.numOutputs)
+    df(params, globals, outputs)
+    outputs(0)
+  }
+
+  def assertWithinEpsilon(x: Double, y: Double): Unit = {
```
andrew-stripe, line 17: Doesn't scalatest have a +- helper for this?

rainier-tests/src/test/scala/com/stripe/rainier/compute/InlineTest.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  check("identity") { x =>
+    x
+  }
+
+  check("LogNormal") { x =>
+    val y = LogNormal(0, 1).param
+    LogNormal(y, y).logDensity(x)
+  }
+}
```
andrew-stripe, line 53: Might be worth testing something with Lookup too since that's the other operation with substantial complexity

andrew-stripe (approved):
Makes sense, some suggestions inline mostly for clarity.

rainier-core/src/main/scala/com/stripe/rainier/core/Sample.scala
avibryant: WAIC requires that we retain the ability to evaluate the likelihood of each individual data point (for whatever definition of "individual data point"). For now that just doesn't seem like a constraint worth adhering to, especially given our discussions about the utility of WAIC-like measures, vs backtesting, in an industry context.

rainier-tests/src/test/scala/com/stripe/rainier/compute/InlineTest.scala
avibryant: LogNormal actually does use Lookup internally (because it has to check that y > 0), which is why I chose it.

rainier-core/src/main/scala/com/stripe/rainier/compute/RealOps.scala
avibryant: Yes, I almost went that way when I initially wrote it, and your suggesting it tips me well over the line.

PR #437, opened 2019-12-27
Move math ops to Decimal

avibryant:
This consolidates all the logic around constant arithmetic ops into Decimal (instead of spreading it between Real and Decimal), including moving Infinity and NegInfinity to be Decimal instances. This is in preparation for giving Placeholder vectors of Decimal and doing partial eval of them.


avibryant:
cc @andrew-stripe for visibility, self -merging

PR #442, opened 2020-01-14
separate out a rainier-compute module

avibryant:
This just moves the `compute` package and all its dependencies to a separate `rainier-compute` sbt module that the `rainier-core` module now depends on, to make the layering more explicit. As a practical matter, this also makes it easier to just compile `compute` while making breaking changes, without seeing the clutter of errors from `core`.


andrew-stripe (approved):
Makes sense to me. I was actually surprised a few times to realize this wasn't a separate module.

PR #444, opened 2020-01-14
allow column vectors as constants in coefficients 

avibryant:
cc @andrew-stripe for visibility

This makes a pair of individually conceptually small (if invasive) changes, but that combine to be quite significant:

- `Coefficients`, used in both `Line` and `LogLine`, now uses `Constant` nodes to hold the constant coefficient values, rather than directly using numeric values
- `Column` is switched from being treated as `NonConstant` to being treated as `Constant`

The overall effect here is that vector data (from using `Fn`) gets propagated through the graph rather than staying as leaf nodes, providing similar pre-computation benefits in the `Fn` case as in the direct case.

We aren't using `Decimal`, so two tests needed to be disabled. The `Decimal` logic will be brought back and folded into `Constant`.


avibryant:
Seems like the lack of inlining is making tests timeout. Will keep this PR unmerged until the follow-ups have resolved that.

PR #454, opened 2020-01-23
Mdoc03

valencik:
Originally from https://github.com/valencik/rainier/pull/1

This PR adds a basic `mdoc` build and example doc file leveraging a few mdoc features, notably plotting via EvilPlot.
Additionally this PR changes the `show` API in `com.stripe.rainier.plot` to return `Drawable`.


avibryant:
Awesome. We'll keep iterating on this but definitely worth merging for now.

PR #459, opened 2020-01-29
Add Scala 2.13.1 build

valencik:
This PR adds 2.13.1 to the cross build.

In doing so it does a number of things. It updates some plugins, Scala versions in build.sbt and Travis, updates sbt, and scalatest.

This might change more things than necessary in the build.sbt.

There was some type resolution funniness about `Parameter` as we have two, one in IR and one in Real, I renamed the IR one to `IParameter` to make things easier on me. We may not want this change permanently.

Currently this PR ignores the `parallel` param to the Model.sample as Scala 2.13 broke out the parallel collections into different library. I don't know if dropping `parallel` is acceptable here.


valencik:
@avibryant Now that this is building successfully in CI we just need confirmation on the `IParameter` change and whether we can drop parallel collection support in `Model.sample`.

If we can drop parallel collection support I'll clean up that area, and if not we'll have to bring in the parallel collections lib.

rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
```diff
@@ -14,7 +14,8 @@ case class Model(private[rainier] val likelihoods: List[Real],
       implicit rng: RNG): Trace = {
     val range =
       if (parallel)
-        1.to(nChains).par
+        // TODO this used to return a ParRange. To do that in 2.13 we would need scala-parallel-collections
+        1.to(nChains)
```
valencik: Decide if we still need `parallel` support and then either clean up or bring in parallel collections.

rainier-compute/src/main/scala/com/stripe/rainier/compute/Evaluator.scala
```diff
@@ -2,6 +2,8 @@ package com.stripe.rainier.compute
 
 class Evaluator(var cache: Map[Real, Double]) extends Numeric[Real] {
 
+  def parseString(str: String): Option[Real] = ???
```
valencik: This is now required by `Numeric`, do we want to just use the underlying `parseString` for `Double` and then create a `Real` from it?

build.sbt
```diff
@@ -6,7 +6,11 @@ lazy val root = project.
   aggregate(rainierBenchmark, rainierTest).
   aggregate(shadedAsm).
   settings(commonSettings).
-  settings(unpublished)
+  settings(unpublished).
+  settings(
+    // crossScalaVersions must be set to Nil on the aggregating project
+    crossScalaVersions := Nil
```
valencik, line 12: This is following the advice of https://www.scala-sbt.org/1.x/docs/Cross-Build.html#Cross+building+a+project

project/plugins.sbt
```diff
@@ -5,6 +5,6 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.1")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.4")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.6")
 addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "1.0.0")
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.6.1")
```
avibryant, line 8: This is unrelated, but while we're in here can we get rid of the code coverage stuff? I find it more annoying than helpful.

rainier-compute/src/main/scala/com/stripe/rainier/compute/Evaluator.scala
```diff
@@ -2,6 +2,10 @@ package com.stripe.rainier.compute
 
 class Evaluator(var cache: Map[Real, Double]) extends Numeric[Real] {
 
+  def parseString(str: String): Option[Real] =
```
avibryant, line 5: The better way to solve this is probably to drop the `extends Numeric[Real]`, and replace references to `Numeric[Real]` elsewhere with a direct reference to `Evaluator`.

rainier-compute/src/main/scala/com/stripe/rainier/ir/IR.scala
```diff
@@ -2,7 +2,7 @@ package com.stripe.rainier.ir
 
 sealed trait Expr
 sealed trait Ref extends Expr
-final class Parameter extends Ref {
+final class IParameter extends Ref {
```
avibryant: Let's use `Param` instead, the same way we use `Const` instead of `Constant`.

rainier-core/src/main/scala/com/stripe/rainier/core/Model.scala
avibryant: We can drop it for now.

avibryant (changes requested):

project/plugins.sbt
avibryant: Also from what I can tell the coverage is what's making CI fail right now.

avibryant (approved):

rainier-compute/src/main/scala/com/stripe/rainier/compute/Evaluator.scala
valencik: note: decided to follow up on this in another PR.

project/plugins.sbt
valencik: decided to hold off on removal for now.

PR #463, opened 2020-01-29
Unify the mdoc and Jupyter plotting support

avibryant:
This PR:

- moves the plotting methods, as well as the println-based convenience methods, to the package object of a new rainier.notebook module
- changes all the plotting methods to return `almond.display.Image` instead of `Drawable`. This should be rendered directly in Jupyter (not yet tested).
- changes the `evilplot` mdoc modified that expected a `Drawable` to instead be `image` and expect a `almond.display.Image`

This lets us use the exact same methods in both contexts.


build.sbt
```diff
@@ -40,8 +40,8 @@ def getPublishTo(snapshot: Boolean) = {
 
 lazy val commonSettings = Seq(
   organization:= "com.stripe",
-  scalaVersion := "2.13.1",
-  crossScalaVersions := List(scalaVersion.value, "2.12.10", "2.11.12"),
+  scalaVersion := "2.12.10",
```
avibryant: This was to make Metals happier, which otherwise was trying to build the plotting code under 2.13

build.sbt
```diff
@@ -111,14 +111,12 @@ lazy val rainierCore = project.
...
   settings(commonSettings)
 
-lazy val rainierPlot = project.
-  in(file("rainier-plot")).
-  settings(name := "rainier-plot").
-  dependsOn(rainierCore).
+lazy val rainierNotebook = project.
+  in(file("rainier-notebook")).
+  settings(name := "rainier-notebook").
   settings(commonSettings).
```
avibryant, line 124: I only needed to delete one method to make rainier-notebook a totally standalone package.

rainier-notebook/src/main/scala/com/stripe/rainier/notebook/package.scala
```diff
@@ -56,24 +55,6 @@ object Jupyter {
   implicit val extent: Extent = Extent(400, 400)
   implicit val theme: Theme = PlotThemes.default
 
-  def trace(t: Trace)(implicit
```
avibryant, line 59: This was not a great implementation of `trace` anyway. We can bring back something better.

build.sbt
valencik: Oh nice, I was encountering this as well.

PR #467, opened 2020-01-31
Ignores metals.sbt

valencik:
As per Metals docs recommendations
https://scalameta.org/metals/docs/editors/eclipse.html#gitignore-projectmetalssbt-metals-and-bloop


avibryant:
👍

PR #471, opened 2020-02-01
add Vec.fill and deprecate param and paramVector

avibryant:
This moves towards a world where `Real(), Vec(), and Generator()` constructors are, consistently, the main way of creating new objects, with appropriate automatic conversions from Distribution etc to each of them.


avibryant:
I think this conflicts with a more important API change, so closing for now.

PR #489, opened 2020-02-25
Metric tuners

avibryant:
This brings back the variance & covariance online estimators as instances of MetricTuner. I still have very low confidence in the dense matrix covariance adaptation; the diagonal matrix adaptation seems more useful, and is set as the default.

Jitter is also added as an option but is disabled by default.


avibryant:
The tests will fail for now because SBC has not been re-run using metric adaptation.

PR #490, opened 2020-03-03
E-BFMI

avibryant:
This adds online computation of the E-BFMI metric, that can be displayed along with other progress information to give a more nuanced view than acceptance rate of how a sampler is doing.

Because BFMI is not a windowed computation, this adds an explicit reset of the stats between the warmup and sampling stages, along with extending Progress to show a message for the current stage.


rainier-sampler/src/main/scala/com/stripe/rainier/sampler/VarianceEstimator.scala
```diff
@@ -0,0 +1,61 @@
+package com.stripe.rainier.sampler
+
+class VarianceEstimator(size: Int) {
```
avibryant, line 3: For BFMI we only use this with size=1, but we'll want this to be multivariate later on for mass matrix adaptation.

PR #501, opened 2021-05-13
change publish to sonatype

kailuowang:
to actually publish, two env variables are needed
"SONATYPE_USERNAME" and "SONATYPE_PASSWORD"
These can be set to a sonatype login token, which can be generated following the [steps here](https://github.com/olafurpg/sbt-ci-release#optional-create-user-tokens)


kailuowang:
these tests seem to be failing prior to these changes in my other PR as well.

avibryant:
@kailuowang those tests are sensitive to RNG. I wonder if maybe the JVM being used to run the tests in CI has changed or and its RNG is different? Or alternatively something around FP math settings?

kailuowang:
Jvm was changed since the old gh action step is no longer supported.  I am using the official Scala action step now. Can we change the tests to expect the new values?

avibryant:
@kailuowang ok, yes, that's a bit of a lengthy process because we have to re-run the Simulation-Based Calibration to make sure the posteriors still look ok, but I have reproduced the test failures by switching to JDK 11 on my machine so I can work on that.

avibryant:
@kailuowang should pass after merging with https://github.com/stripe/rainier/pull/502

kailuowang:
@avibryant I merged in your commits, the build passes, but the CLAssistant got confused.

